PR #31, opened 2022-02-16
Confluence/remove grpc

confluence:
This is an update of the wrapper to use the new HTTP scripting API.

This PR should be merged at the same time as the companion PRs in the [backend](https://github.com/CARTAvis/carta-backend/pull/1022), [frontend](https://github.com/CARTAvis/carta-frontend/pull/1785), [protocol buffers](https://github.com/CARTAvis/carta-protobuf/pull/65), [controller](https://github.com/CARTAvis/carta-controller/pull/110) and [website](https://github.com/CARTAvis/CARTAvis.github.io/pull/18).

### DIFF
--- a/carta/protocol.py
+++ b/carta/protocol.py
@@ -25,9 +25,22 @@ class Protocol:
     frontend_url : string
         The URL of the frontend.
     token : :obj:`carta.token.Token`, optional
-        The security token used by the CARTA instance. May be omitted if the URL contains a token.
+        The security token used by the CARTA instance. May be omitted if the URL contains a token. If this is a controller token, it will be assumed to be a refresh token, and controller authentication will be enabled. If it's a backend token, backend authentication will be enabled.
     debug_no_auth : boolean
         Disable authentication. This should be set if the backend has been started with the ``--debug_no_auth`` option. This is provided for debugging purposes only and should not be used under normal circumstances.
+
+    Attributes
+    ----------
+    frontend_url : string
+        The URL of the frontend.
+    base_url : string
+        The base frontend URL (with the token parameter, if any, removed).
+    auth : integer
+        The authentication type. One of :obj:`carta.protocol.AuthType.BACKEND`,  :obj:`carta.protocol.AuthType.CONTROLLER` or  :obj:`carta.protocol.AuthType.NONE`.
+    refresh_token : :obj:`carta.token.ControllerToken`, optional
+        The controller refresh token. Will be set to the ``token`` parameter if it is a controller token.
+    backend_token : :obj:`carta.token.BackendToken`, optional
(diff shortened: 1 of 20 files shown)


carta/protocol.py
```diff
@@ -0,0 +1,341 @@
...
+
+        token = ControllerToken(token_string)
+
+        if path:
+            token.save(path)
+
+        return token
+
+    @classmethod
+    def split_token_from_url(cls, url):
```
YuHsuan-Hwang: shall we change it to a static method if `cls` is not used in the method?

carta/session.py
```diff
@@ -0,0 +1,619 @@
...
+    ----------
+    session_id : integer
+        The ID of the CARTA frontend session associated with this object. This is set automatically when a new session is created with :obj:`carta.session.Session.create` or :obj:`carta.session.Session.start_and_create`.
+    protocol : :obj:`carta.protocol.Protocol`
+        The protocol object which handles HTTP communication with the CARTA scripting API. This is created automatically when a new session is created with one of the three class methods for creating sessions.
+    browser : :obj:`carta.browser.Browser`
+        The browser object associated with this session. This is created automatically when a new session is created with :obj:`carta.session.Session.create` or :obj:`carta.session.Session.start_and_create`.
+    backend : :obj:`carta.backend.Backend`
+        The backend object associated with this session. This is created automatically when a new session is created with :obj:`carta.session.Session.start_and_create`.
+    debug_no_auth : boolean
```
YuHsuan-Hwang: may remove `debug_no_auth` from the description. I found that some class descriptions are listing the attributes, but some are not. Do we in general list them all, or just list them when further description is necessary?

carta/protocol.py
confluence: Good idea.

carta/session.py
confluence: Ah, that's left over from when it was a parameter to this class. I will remove it.

I haven't been very consistent about listing attributes (since they're often the same as the parameters); I'll go through the classes and add them in. I believe that the convention is to list all public attributes (so not the ones with leading underscores which are intended to be private).

YuHsuan-Hwang: I see. maybe we can shorten the descriptions when they're the same as the parameters

confluence: That's actually less of a problem than I thought. I definitely needed to document the `Protocol` attributes!

I couldn't find a way to make Sphinx's autodoc integrate properties into the class docstring as attributes, so I documented them as if they were methods, and added return types to all of them. I made a small change to the functionality of `Image.header`, which now returns more processed output rather than the raw header data from the frontend. I'm thinking of also converting timestamps to dates automatically, but that may be too magical.

YuHsuan-Hwang (approved):

PR #45, opened 2022-09-06
Add fallback for entries with no values; aggregate HISTORY, COMMENT and blank entries

### LINKED ISSUE #44: cannot get image header due to HISTORY
```
    @cached
    def header(self):
        """The header of the image.

        Returns
        -------
        dict of string to string, integer, float or boolean
            The header of the image, with field names as keys. T or F string values are automatically converted to booleans.
        """
        raw_header = self.get_value("frameInfo.fileInfoExtended.headerEntries")
        header = {e["name"]: e.get("numericValue", e["value"]) for e in raw_header}
        for k, v in header.items():
            if v == 'T':
                header[k] = True
            elif v == 'F':
                header[k] = False
        return header
```
of the `Image` class needs a fix for the HISTORY in the `raw_header`.

The example `raw_header` causing the issue is like

```
{'entryType': 'FLOAT', 'name': 'OBSGEO-Z', 'numericValue': -2481029.851874, 'value': '-2.481029851874E+06'}
{'comment': 'Date FITS file was written', 'name': 'DATE', 'value': '2019-04-02T02:49:20.966000'}
{'name': 'ORIGIN', 'value': 'CASA 5.3.0-143'}
{'name': 'HISTORY CASA START LOGTABLE'}
```

as there is only 'name' but no 'value' in `{'name': 'HISTORY CASA START LOGTABLE'}`.

trackback
```
  File "/Users/kswang/carta_pythonInterface_testing/test.py", line 6, in <module>
    print(img0.header)
  File "/opt/anaconda3/lib/python3.9/site-packages/carta/util.py", line 113, in newfunc
    self._cache[func.__name__] = func(self, *args)
  File "/opt/anaconda3/lib/python3.9/site-packages/carta/image.py", line 163, in header
    header = {e["name"]: e.get("numericValue", e["value"]) for e in raw_header}
  File "/opt/anaconda3/lib/python3.9/site-packages/carta/image.py", line 163, in <dictcomp>
    header = {e["name"]: e.get("numericValue", e["value"]) for e in raw_header}
KeyError: 'value'
```

confluence:
Fixes #44. What was initially a simple fix became more complex when I realised that `HISTORY` entries can be duplicates, which means that they can't be used as dictionary keys (and it's also not very useful conceptually to do this). I have done something similar to [what Astropy does](https://docs.astropy.org/en/stable/io/fits/usage/headers.html#comment-history-and-blank-keywords) and aggregated `HISTORY`, `COMMENT` and blank keyword entries into single entries with list values. Additionally, I use the `>` continuation marker in the `HISTORY` entries to concatenate wrapped lines back together.

I have kept the `None` fallback in case we encounter other entries with no values (I don't know if that's legal FITS, but if it isn't, that doesn't mean someone isn't going to do it).

### DIFF
--- a/carta/image.py
+++ b/carta/image.py
@@ -155,7 +155,7 @@ class Image:
             The header of the image, with field names as keys. T or F string values are automatically converted to booleans.
         """
         raw_header = self.get_value("frameInfo.fileInfoExtended.headerEntries")
-        header = {e["name"]: e.get("numericValue", e["value"]) for e in raw_header}
+        header = {e["name"]: e.get("numericValue", e.get("value", None)) for e in raw_header}
         for k, v in header.items():
             if v == 'T':
                 header[k] = True
--- a/carta/image.py
+++ b/carta/image.py
@@ -149,10 +149,16 @@ class Image:
     def header(self):
         """The header of the image.
 
+        Entries with T or F string values are automatically converted to booleans.
+
+        ``HISTORY``, ``COMMENT`` and blank keyword entries are aggregated into single entries with list values and with ``'HISTORY'``, ``'COMMENT'`` and ``''`` as keys, respectively. An entry in the history list which begins with ``'>'`` will be concatenated with the previous entry. Adjacent ``COMMENT`` entries will be concatenated.
+
+        Any other header entries with no values are given values of ``None``.
+
         Returns
         -------
         dict of string to string, integer, float, boolean, ``None`` or list of strings
-            The header of the image, with field names as keys. T or F string values are automatically converted to booleans. ``HISTORY``, ``COMMENT`` and blank keyword entries are aggregated into single entries with list values and with ``'HISTORY'``, ``'COMMENT'`` and ``''`` as keys, respectively. An entry in the history list which begins with ``'>'`` will be concatenated with the previous entry. Any other header entries with no values are given values of ``None``.
+            The header of the image, with field names as keys.
         """
         raw_header = self.get_value("frameInfo.fileInfoExtended.headerEntries")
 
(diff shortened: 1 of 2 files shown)


kswang1029 (approved):
this fix works fine. 👍 

one minor issue about COMMENT but this is not that important for the scripting interface so please consider this just as a note.

Adjacent COMMENTs should be considered as one comment. Unfortunately there seems no such special char like ">" for HISTORY as an indicator of multiple lines.

Screen Shot 2022-09-07 at 14 00 36

So currently, COMMENTs may be like the following

```
COMMENT ['This FITS file may contain long string keyword values that are', "continued over multiple keywords.  This convention uses the  '&'", 'character at the end of a string which is then continued', "on subsequent keywords whose name = 'CONTINUE'.", 't=Scalar', 't=Scalar']
```

confluence:
OK, I think I can fix that by keeping track of the last comment position.

confluence:
@kswang1029 please re-review.

There are still a few minor issues, e.g. some common comments (but not all!) are indented by an extra space; I could post-process the comments by stripping spaces off the ends and removing double spaces inside, but this wouldn't help if a comment were indented *and* split in the middle of a word. Some comments are also positional labels describing other entries, so they lose meaning if they're aggregated. I don't think this is a huge issue for scripting; the user has the option of bypassing this function and getting the raw header data from the frontend.

Ideally I would also like to offer the original string header (or at least the same reconstruction we show in the GUI) -- if there's no way to fetch that directly, I could reconstruct it from the raw data. Do you think this is important to add now? If not, we could merge this as-is and create a separate issue.

kswang1029:
let's get this merge now :)

kswang1029:
> let's get this merge now :)

For the future, we can add 'RAW_HEADER' in the dict with value as a well formatted text like the header viewer of ds9. Low priority though.

PR #52, opened 2022-09-14
Confluence/polarization

### LINKED ISSUE #51: extend Image.set_channel_stokes for computed polarization components
The "Stokes" slider in the animator widget has been generalized to "Polarization" to include other non-Stokes polarization components. The full "Stokes" enums (table 29 of https://fits.gsfc.nasa.gov/standard40/fits_standard40aa-le.pdf) is

Screen Shot 2022-09-08 at 09 30 44

In addition to the above components, CARTA internally generates "computed" components from Stokes IQUV including
- Ptotal
- Plinear
- PFtotal
- PFlinear
- Pangle
The exact list of the computed components depends on the integrity of Stokes IQUV. For example, if Stokes I is missing, fractional ones are not included. The protobuf enums can be found at https://carta-protobuf.readthedocs.io/en/latest/enums.html#polarizationtype

Screen Shot 2022-09-08 at 09 41 52

I suggest we also change `set_channel_stokes` to `set_channel_polarization` for consistency. Relevant "stokes" variables in the codebase may need to be revised too for better readability. 

One minor issue about the validation of the stokes:
if we load a IQUV cube and do
`img0.set_channel_stokes(55, 4, recursive=True)`
an exception occurs
```
  File "/opt/anaconda3/lib/python3.9/site-packages/carta/validation.py", line 615, in newfunc
    param.validate(value, self)
  File "/opt/anaconda3/lib/python3.9/site-packages/carta/validation.py", line 559, in validate
    param.validate(value, parent)
  File "/opt/anaconda3/lib/python3.9/site-packages/carta/validation.py", line 152, in validate
    raise ValueError(f"{value} is greater than or equal to maximum value {self.max}.")
ValueError: 4 is greater than or equal to maximum value 4.

During handling of the above exception, another exception occurred:

  File "/Users/kswang/carta_pythonInterface_testing/test.py", line 13, in <module>
    img0.set_channel_stokes(55, 4, recursive=True)
  File "/opt/anaconda3/lib/python3.9/site-packages/carta/validation.py", line 628, in newfunc
    raise CartaValidationFailed(f"Invalid function parameter: {msg}")
carta.util.CartaValidationFailed: Invalid function parameter: 4 is greater than or equal to maximum value 4.
```
I think this is the classic 0-based/1-based index issue. `ValueError: 4 is greater than or equal to maximum value 4.` should be `ValueError: 4 is greater than or equal to maximum value 3.` (NOTE: here the computed components are not considered. This is what it does in the current codebase).

confluence:
Fixes #51. Any available polarization can be set (always by its numeric value). Keyword aliases are provided through a helper class. A list of available polarizations can be retrieved, and it can be queried by keyword. `polarizations` includes computed polarizations; `num_polarizations` does not.

Examples:

    from carta.constants import Polarization
    img.polarizations
    img.num_polarizations
    img.has_polarization(Polarization.Q)
    img.set_channel_polarization(polarization=Polarization.Q)
    img.set_channel_polarization(polarization=2)
    img.set_channel_polarization(polarization=Polarization.PANGLE)
    img.set_channel_polarization(polarization=17)

Future enhancement (in a separate issue): add a wrapper for return values (like `polarizations`) so that they are displayed as keywords rather than literals. For now the user will see the underlying integer values.

Possible future enhancement: set a (non-computed) polarization by index? Would this be useful?

### DIFF
--- a/carta/constants.py
+++ b/carta/constants.py
@@ -113,3 +113,24 @@ class ContourDashMode:
     NONE = "None"
     DASHED = "Dashed"
     NEGATIVE_ONLY = "NegativeOnly"
+
+
+class Polarization:
+    """Polarizations, corresponding to the POLARIZATIONS enum in the frontend."""
+    YX = -8
+    XY = -7
+    YY = -6
+    XX = -5
+    LR = -4
+    RL = -3
+    LL = -2
+    RR = -1
+    I = 1
+    Q = 2
+    U = 3
+    V = 4
+    PTOTAL = 13
+    PLINEAR = 14
+    PFTOTAL = 15
+    PFLINEAR = 16
+    PANGLE = 17
--- a/carta/image.py
+++ b/carta/image.py
@@ -342,7 +342,7 @@ class Image:
 
     # NAVIGATION
 
-    @validate(Evaluate(Number, 0, Attr("depth"), Number.INCLUDE_MIN), Evaluate(Number, 0, Attr("num_stokes"), Number.INCLUDE_MIN), Boolean())
+    @validate(Evaluate(Number, 0, Attr("depth"), Number.INCLUDE_MIN), Evaluate(Number, 1, Attr("num_stokes")), Boolean())
     def set_channel_stokes(self, channel=None, stokes=None, recursive=True):
         """Set the channel and/or Stokes.
 
--- a/carta/image.py
+++ b/carta/image.py
@@ -4,9 +4,9 @@ Image objects should not be instantiated directly, and should only be created th
 """
 import posixpath
 
-from .constants import Colormap, Scaling, SmoothingMode, ContourDashMode
+from .constants import Colormap, Scaling, SmoothingMode, ContourDashMode, Polarization
 from .util import Macro, cached
-from .validation import validate, Number, Color, Constant, Boolean, NoneOr, IterableOf, Evaluate, Attr
+from .validation import validate, Number, Color, Constant, Boolean, NoneOr, IterableOf, Evaluate, Attr, Attrs, OneOf
 
 
 class Image:
@@ -227,7 +227,7 @@ class Image:
             The shape of the image; dimensions ordered with width last.
 
         """
-        return list(reversed([self.width, self.height, self.depth, self.num_stokes][:self.ndim]))
+        return list(reversed([self.width, self.height, self.depth, self.num_polarizations][:self.ndim]))
 
     @property
     @cached
@@ -267,13 +267,13 @@ class Image:
 
     @property
     @cached
-    def num_stokes(self):
-        """The number of Stokes parameters of the image.
+    def num_polarizations(self):
+        """The number of polarizations of the image, excluding computed polarizations.
 
         Returns
         -------
         integer
-            The number of Stokes parameters.
+            The number of polarizations.
         """
         return self.get_value("frameInfo.fileInfoExtended.stokes")
 
@@ -288,6 +288,33 @@ class Image:
             The number of dimensions.
         """
         return self.get_value("frameInfo.fileInfoExtended.dimensions")
+
+    @property
+    @cached
+    def polarizations(self):
+        """The available polarizations of the image.
+
+        This includes Stokes parameters, correlations, and computed components.
+
+        Returns
+        -------
+        list of integers
+            The available polarizations.
+        """
+        return self.get_value("polarizations")
+
+    @validate(Constant(Polarization))
+    def has_polarization(self, polarization):
+        """Check whether a polarization is available.
+
+        This includes Stokes parameters, correlations, and computed components.
+
+        Returns
+        -------
+        boolean
+            Whether the polarization is available.
+        """
+        return polarization in self.polarizations
 
     # SELECTION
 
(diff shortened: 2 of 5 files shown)


kswang1029 (changes requested):
when we do `print(img0.polarizations)` and see `[1, 2, 3, 4, 13, 14, 15, 16, 17]`, then if we do `print(img0.has_polarization(13))`, it is True returned. But if we do `print(img0.has_polarization(12))`, we will see

> Traceback (most recent call last):
>   File "/opt/anaconda3/lib/python3.9/site-packages/carta/validation.py", line 624, in newfunc
>     param.validate(value, self)
>   File "/opt/anaconda3/lib/python3.9/site-packages/carta/validation.py", line 254, in validate
>     raise ValueError(f"{value} is not {self.description}")
> ValueError: 12 is not a class property of :obj:`carta.constants.Polarization`
> 
> During handling of the above exception, another exception occurred:
> 
> Traceback (most recent call last):
>   File "/Users/kswang/carta_pythonInterface_testing/test.py", line 18, in <module>
>     print(img0.has_polarization(12))
>   File "/opt/anaconda3/lib/python3.9/site-packages/carta/validation.py", line 636, in newfunc
>     raise CartaValidationFailed(f"Invalid function parameter: {msg}")
> carta.util.CartaValidationFailed: Invalid function parameter: 12 is not a class property of carta.constants.Polarization

instead of a False.

confluence:
This is by design -- there is no polarization with value 12 (or 0, or 12345). So you can't check any of those, because they aren't valid polarizations. The input to the function is validated to disallow anything except a polarization value.

I could change this behaviour to allow any number to be checked, but I think that would be less correct (the purpose of validating the input is to prevent the user from entering an invalid integer value by mistake).

kswang1029:
> This is by design -- there is no polarization with value 12 (or 0, or 12345). So you can't check any of those, because they aren't valid polarizations. The input to the function is validated to disallow anything except a polarization value.
> 
> I could change this behaviour to allow any number to be checked, but I think that would be less correct (the purpose of validating the input is to prevent the user from entering an invalid integer value by mistake).

I see. My concern here is `Image.has_polarization()` returns True or raises an exception. Not sure what application of this function would be? Maybe we don't need `Image.has_polarization()` at all as `Image.polarizations` can already tell users the available polarizations as a list. 🤔

kswang1029:
> > This is by design -- there is no polarization with value 12 (or 0, or 12345). So you can't check any of those, because they aren't valid polarizations. The input to the function is validated to disallow anything except a polarization value.
> > I could change this behaviour to allow any number to be checked, but I think that would be less correct (the purpose of validating the input is to prevent the user from entering an invalid integer value by mistake).
> 
> I see. My concern here is `Image.has_polarization()` returns True or raises an exception. Not sure what application of this function would be? Maybe we don't need `Image.has_polarization()` at all as `Image.polarizations` can already tell users the available polarizations as a list. 🤔

@confluence never mind. I think we can keep `Image.has_polarization()` as using constants (eg Polarization.Q) as input is still useful.

confluence:
The main purpose is to allow the user to look up a polarization by keyword name. `polarizations` is a list of numbers. I added an issue (#53) for this -- I think it should be possible to create wrappers for this data so that the values are displayed as keywords (or possibly there's an existing enum-like feature in Python that I could use). If I put that in, it would mostly eliminate the need for functions like this.

kswang1029:
For an image having 480 channels (index 0~479), if we do `img1.set_channel(480)`, we will see

> carta.util.CartaValidationFailed: Invalid function parameter: 480 is greater than or equal to maximum value 480.


it should be 

> carta.util.CartaValidationFailed: Invalid function parameter: 480 is greater than or equal to maximum value 479.

confluence:
Ah, good point. ~It should actually be "greater than maximum value of 480".~ I need to revise this wording (I can't just increment / decrement the bound because it also needs to work if the numbers are floats).

confluence:
The message is currently technically correct; it just isn't conveying the information that the upper bound may not be included. I'll start by changing minimum / maximum value to lower / upper bound, since the current phrasing implies that the bound is included. I have an idea for the rest.

confluence:
Upon reflection, I have also removed the `has_polarization` function, since the user can already use a keyword in exactly the same way to check whether `Polarization.I in img.polarizations` directly. I don't know what I was thinking before.

kswang1029:
uh, it seems the revision breaks set_channel. In the following example, channel 20 is a valid one.

> img1.set_channel(20)
  File "/opt/anaconda3/lib/python3.9/site-packages/carta/validation.py", line 630, in newfunc
    param.validate(value, self)
  File "/opt/anaconda3/lib/python3.9/site-packages/carta/validation.py", line 574, in validate
    param.validate(value, parent)
  File "/opt/anaconda3/lib/python3.9/site-packages/carta/validation.py", line 158, in validate
    raise ValueError(f"{value} is {cmp_str} upper bound {self.max}, but must be smaller.")
UnboundLocalError: local variable 'cmp_str' referenced before assignment

confluence:
Should be fixed now.

kswang1029 (approved):
this looks good now. 👍

PR #67, opened 2023-01-09
Use a conditional import for Hashable for compatibility with Python >= 3.10

### LINKED ISSUE #66: "Hashable" class change for Python 3.10+
If using Python 3.10, we get the following error: ```ImportError: cannot import name 'Hashable' from 'collections'```.

I found the solution [here](https://bobbyhadz.com/blog/python-importerror-cannot-import-name-hashable-from-collections). We just need to modify line 6 of `carta/validation.py` from:

> from collections import Hashable

to

> from collections.abc import Hashable

confluence:
Fixes #66.

### DIFF
--- a/carta/validation.py
+++ b/carta/validation.py
@@ -3,7 +3,11 @@
 import re
 import functools
 import inspect
-from collections import Hashable
+
+try:
+    from collections.abc import Hashable
+except ImportError:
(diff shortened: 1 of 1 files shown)


ajm-ska (approved):
I confirm that this works on Python 3.10.

Issue #69, opened 2023-02-22
Function to start backend and connect interactively to default session created in user's browser


confluence:
As requested by @Jordatious . This should also work with the Electron app. Only for local executables (does not apply to the controller).

This requires a bit of glue, but mostly reuses existing functionality.

PR #70, opened 2023-02-28
Functionality for starting the backend and connecting to the default (interactive) session

confluence:
Implements #69. Note that this is new behaviour which is distinct from starting the backend and creating *headless* sessions -- it allows the user to start a new interactive CARTA session from inside a Jupyter notebook, Python interpreter, etc., instead of starting CARTA separately and copying the frontend URL and session ID into the `interact` method by hand. This is not intended for use from non-interactive scripts -- that doesn't really make sense because the backend process is killed when the Python session exits (if you do this by mistake you should be able to start CARTA again and reconnect from your orphaned browser session; this seemed to work when I tried it).

I have tested this on Ubuntu. It still needs to be tested with the Electron app.

This *should* work with the Electron executable as long as it
1. accepts the `--enable_scripting` flag
2. prints the full backend output to the terminal (including the frontend URL and session ID lines).
 
It should also work with the launcher script used to bypass Electron for the headless session, assuming that it meets the same requirements and also does not suppress the opening of a new session in the browser by default. (It should, however, accept the `--no_browser` flag, so that it can do this when it is used for headless sessions).

Usage:

```
from carta.session import Session
session = Session.start_and_interact()
```

By default this method assumes that the executable name / path is just `carta`; to specify a different executable path, use the `executable_path` parameter.

### DIFF
--- a/carta/backend.py
+++ b/carta/backend.py
@@ -25,6 +25,8 @@ class Backend:
         If this is set, an attempt will be made to start the backend over ``ssh`` on this host.
     token : :obj:`carta.token.BackendToken`
         If this is set, this will be used as the security token and no attempt will be made to parse the token from the backend output.
+    session_creation_timeout : integer
+        How long to keep checking the output for a default session ID. If this is set to zero (which is the default), no attempt is made to parse a session ID from the output. The calling function should set this to a non-zero value if parsing the session ID is required.
 
     Attributes
     ----------
@@ -40,6 +42,10 @@ class Backend:
         All output of the backend process, split into lines, terminated by newline characters.
     errors : list of strings
         Error output of the backend process, split into lines, terminated by newline characters.
+    last_session_id : integer
+        The ID of the last session connected to this backend process, parsed from the process output. This is likely to be the default session automatically created in the user's browser on startup, if this functionality was not suppressed with the ``--no_browser`` flag. This value is used by the :obj:`carta.session.Session.start_and_interact` method, which connects to this default session. It is not used by the session creation methods which use a wrapper-controlled headless browser, as those parse the session ID from the browser session.
+    session_creation_timeout : integer
+        How long to keep checking the output for a default session ID. If this is set to zero, no attempt is made to parse a session ID from the output.
 
     Returns
     -------
@@ -49,17 +55,31 @@ class Backend:
     """
     FRONTEND_URL = re.compile(r"CARTA is accessible at (http://.*?:\d+/\?token=(.*))")
     FRONTEND_URL_NO_AUTH = re.compile(r"CARTA is accessible at (http://(.*?):\d+.*)")
+    SESSION_ID = re.compile(r"Session (\d+) \[[\d.]+\] Connected.")
 
-    def __init__(self, params, executable_path="carta", remote_host=None, token=None):
+    def __init__(self, params, executable_path="carta", remote_host=None, token=None, session_creation_timeout=0):
         self.proc = None
         self.frontend_url = None
         self.token = token
         self.debug_no_auth = ("--debug_no_auth" in params)
         self.output = []
         self.errors = []
+        self.last_session_id = None
+        self.session_creation_timeout = session_creation_timeout
 
         ssh_cmd = ("ssh", "-tt", remote_host) if remote_host is not None else tuple()
         self.cmd = tuple(str(p) for p in (*ssh_cmd, executable_path, *params))
+
+    def update_output(self):
+        while True:
+            line = self.proc.stdout.readline()
+            if not line:
+                break
+            line = line.decode()
+
+            self.output.append(line)
+            if "[error]" in line or "[critical]" in line:
+                self.errors.append(line)
 
     def start(self):
         """Start the backend process.
@@ -71,16 +91,7 @@ class Backend:
         os.set_blocking(self.proc.stdout.fileno(), False)
 
         time.sleep(1)
-
-        while True:
-            line = self.proc.stdout.readline()
-            if not line:
-                break
-            line = line.decode()
-
-            self.output.append(line)
-            if "[error]" in line or "[critical]" in line:
-                self.errors.append(line)
+        self.update_output()
 
         if self.proc.poll() is not None:
             return False
@@ -95,6 +106,25 @@ class Backend:
 
         if self.token is None and not self.debug_no_auth:
             self.token = BackendToken(token_string)
+
+        # Only try to parse the session ID if it has been requested
+        if self.session_creation_timeout > 0:
+            start = time.time()
+
+            while self.last_session_id is None:
+                if time.time() - start > self.session_creation_timeout:
+                    break
+
+                # Check for new output
+                self.update_output()
+
+                for line in self.output:
+                    m = self.SESSION_ID.search(line)
+                    if m:
+                        self.last_session_id = m.group(1)
+                        break
+
+                time.sleep(1)
 
         return True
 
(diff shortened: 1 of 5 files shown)


ajm-ska:
1. 
This PR seems to be using an older version of `carta/validation.py` because I get:   

> File "/Users/ajm/carta-python/carta/validation.py", line 6, in <module>
>     from collections import Hashable
> ImportError: cannot import name 'Hashable' from 'collections' (/opt/homebrew/Cellar/python@3.11/3.11.2_1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/collections/__init__.py)

That issue was fixed 2 months ago in the 'dev' branch.


2. 
Is there a way for it to install the selenium requirement automatically in the `pip install .` step? Maybe adding it to the `carta.egg-info/requires.txt` list? Otherwise I get:

>   File "/Users/ajm/carta-python/carta/browser.py", line 5, in <module>
>     from selenium import webdriver
> ModuleNotFoundError: No module named 'selenium'.

and need to do `pip3 install selenium`.


3.
It works fine when using the Electron `carta.sh` bypass script:

> session = Session.start_and_interact(executable_path="/Applications/CARTA.app/Contents/Resources/app/carta-backend/bin/carta.sh")

However, if just starting the Electron version directly with:

> session = Session.start_and_interact(executable_path="/Applications/CARTA.app/Contents/MacOS/CARTA")

I immediately get the following error:

> Traceback (most recent call last):
>   File "/Users/ajm/carta-python/test.py", line 10, in <module>
>     session = Session.start_and_interact(executable_path="/Applications/CARTA.app/Contents/MacOS/CARTA")
>               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/ajm/carta-python/carta/session.py", line 131, in start_and_interact
>     if not backend.start():
>            ^^^^^^^^^^^^^^^
>   File "/Users/ajm/carta-python/carta/backend.py", line 108, in start
>     self.token = BackendToken(token_string)
>                               ^^^^^^^^^^^^
> UnboundLocalError: cannot access local variable 'token_string' where it is not associated with a value

and then the Electron window opens.

If I run `/Applications/CARTA.app/Contents/MacOS/CARTA` in the terminal, I do see the full carta_backend terminal output with the Frontend URL and Session ID lines. 

It feels like the Python script immediately finishes before the carta_backend process has a chance to start up. So maybe the default `session_creation_timeout=10` added to "start_and_interact" is not working?

I know this is not intended for non-interactive scripts, but what I was intending to do was just open a file and save it:

> from carta.session import Session
> from carta.token import ControllerToken
> 
> from carta.session import Session
> from carta.token import BackendToken
> import numpy as np
> from carta.constants import Colormap, Scaling
> 
> session = Session.start_and_interact(executable_path="/Applications/CARTA.app/Contents/MacOS/CARTA", session_creation_timeout=20) 
> img = session.open_image("/Users/ajm/CARTA/Images/HLTau_Band7_Continuum.fits")
> session.save_rendered_view("test.png", "white")

So I could simply use the default web-browser rather than Electron for this, but is it supposed wait 10 seconds?

4.
Another ideal use case for this functionality is with the Linux AppImage! It would be great for users on Linux who want to use python scripting but without root access to install CARTA.

confluence:
1. Oops; I think I created the branch from a local dev checkout which hadn't been updated. I'll fix that.
2. Selenium is intended to be an optional dependency (since it's required only for headless sessions), but it looks like the import may be happening even if you aren't using a headless session -- I'll check that.
3. You can override the session start timeout, but that error doesn't have anything to do with the session ID -- there's a problem with the token that is parsed from the URL, and I may be overlooking something that the Electron app is doing differently. Could you paste the full terminal output from the Electron app? Have we confirmed that the URL printed by the Electron app can be used with the manual interactive case?
4. Yes, this should be usable with any CARTA executable. But all the other invocation methods should also work with the AppImage, surely.

confluence:
@ajm

1. I have merged in the update.
2. I have checked, and using this method should not require importing `browser` at any point. Are you importing it elsewhere in your test script?

ajm-ska:
@confluence Thanks for checking. For No.2 it looks like I had `from carta.browser import Chrome` in a script.

As for No.3, here is the terminal output if I run the Electron version directly in the terminal:

> ajm@almac ~ % /Applications/CARTA.app/Contents/MacOS/CARTA
> CARTA will use the default ephemerides and geodetic data.
> 
> [2023-03-01 07:36:50.031] [CARTA] [info] Writing to the log file: /Users/ajm/.carta/log/carta.log
> [2023-03-01 07:36:50.031] [CARTA] [info] /Applications/CARTA.app/Contents/Resources/app/carta-backend/bin/carta_backend: Version 3.0.0
> 
> [2023-03-01 07:36:50.032] [CARTA] [info] Serving CARTA frontend from /Applications/CARTA.app/Contents/Resources/app
> 
> [2023-03-01 07:36:50.034] [CARTA] [info] Listening on port 35629 with top level folder /, starting folder /Users/ajm. The number of OpenMP worker threads will be handled automatically.
> [2023-03-01 07:36:50.034] [CARTA] [info] CARTA is accessible at http://localhost:35629/?token=8af99761-6a4f-44b6-b4aa-d5aaece64f83
> 
> [2023-03-01 07:36:54.146] [CARTA] [info] 0x7f97fdf04080 ::Session (3333622949:1)
> [2023-03-01 07:36:54.146] [CARTA] [info] Session 3333622949 [127.0.0.1] Connected. Num sessions: 1

I have just checked. I have a `~/.carta/backend.json` containing: 
```
{
   "enable_scripting": true
}
```
and if I use "Session.interact", I confirm that it can be used to open an image in the running Electron window:

> session = Session.interact("http://localhost:35629/?token=8af99761-6a4f-44b6-b4aa-d5aaece64f83", 3333622949)

It still feels to me that "Session.start_and_interact" is happening too fast. I get the Python error immediately, a few seconds before the Electron window appears. In the Electron start-up script, Electron opens the window first, and then a child process starts up the carta_backend. So I think it needs to wait a bit for the carta_backend to start up.

confluence:
Oh, I think I know what may be happening -- the initial output may be taking too long to appear, which may be causing the code for parsing the output to exit before any output has been printed. I'll tweak this to use an explicit timeout.

confluence:
@ajm please try again. You should now also get a better error if the URL isn't found (previously this failure wasn't being handled correctly).

Please also make sure that this works if the scripting is *not* turned on by default in the configuration file -- the wrapper uses the `--enable_scripting` flag, so it needs to be handled by both of the executables.

ajm-ska (approved):
@confluence This works perfect now! In my basic test script, it opens the Electron window, opens an image, saves it, and then closes the Electron window.

It also works without the "enable_scripting" setting in `~/.carta/backend.json`.

So it all seems fine to me at least. @Jordatious and @kswang1029 will probably be able to test more advanced usage.

Jordatious:
@confluence and I have discussed this a bit. It's working well for me when I launch the Electron app directly from the scripting interface by using the Electron executable like so:

```
session = Session.start_and_interact(executable_path='/Applications/CARTA.app/Contents/MacOS/CARTA')
```

If I set `carta` as an alias, as suggested [here](https://cartavis.org/#download), and don't pass `executable_path` in, it doesn't work as expected. I have to actually create an executable `carta` script in my PATH like so

```
#!/bin/bash

/Applications/CARTA.app/Contents/MacOS/CARTA
```

Like previously, I can also use Chrome to launch CARTA (e.g. headless), but I need to pass in the bypass executable like so (otherwise both the Electron app and Chrome open simultaneously, although only the latter is used):

 ```
 session = Session.start_and_create(Chrome(headless=headless,options=chromeOptions), '/Applications/CARTA.app/Contents/Resources/app/carta-backend/bin/carta.sh')
 ```
 
 That's a little fiddly and unintuitive for other users, but perhaps we can document that somewhere.

Jordatious (approved):

Issue #55, opened 2022-09-14
add Image.set_cube_matching as a shortcut to match image both spatially and spectrally


kswang1029:
add `Image.set_cube_matching` as a shortcut. Effectively, it does `Image.set_spatial_matching` and `Image.set_spectral_matching` together with the same input boolean. 

having a new `Image.make_cube_reference` as a shortcut of `Image.make_spatial_reference` and `Image.make_spectral_reference` is also nice to UX I think.

confluence:
Should these also match the raster scaling / set the raster scaling reference? I can see the utility of functions that group all three settings, but I'm less convinced about functions that group 2 out of 3.

kswang1029:
> Should these also match the raster scaling / set the raster scaling reference? I can see the utility of functions that group all three settings, but I'm less convinced about functions that group 2 out of 3.

I suggest we only group spatial and spectral matching. 

Well if you like we can also make a matching everything function. Then users can decide which one to use for their various use cases.

PR #73, opened 2023-03-03
Add a convenience function to set spatial and spectral matching together.

confluence:
Implements #55.

@kswang1029 does it make sense to add a corresponding `make_cube_reference` function to make an image the spatial and spectral reference at the same time?


kswang1029 (approved):
👍

Issue #62, opened 2022-10-03
create a high-level wrapper to plot contours


kswang1029:
based on the existing functions
- Image.configure_contours()
- Image.set_contour_dash()
- Image.set_contour_color()
- Image.set_contour_colormap()
- Image.apply_contours()

we can actually provide a high-level wrapper of these to a single function such as Image.plot_contours(). This would simplify the typesetting of a script probably. It would be more intuitive for users to plot contours without calling all those fine steps.

PR #74, opened 2023-03-03
Add a helper function for configuring all contour properties in one step

confluence:
Implements #62.

### DIFF
--- a/carta/image.py
+++ b/carta/image.py
@@ -587,32 +587,25 @@ class Image:
             The smoothing mode.
         smoothing_factor : {2}
             The smoothing factor.
-        dash_mode : {0}
+        dash_mode : {3}
             The dash mode.
-        thickness : {1}
+        thickness : {4}
             The dash thickness.
-        color : {0}
+        color : {5}
             The color.
-        colormap : {0}
+        colormap : {6}
             The colormap.
-        bias : {1}
+        bias : {7}
             The colormap bias.
-        contrast : {2}
+        contrast : {8}
             The colormap contrast.
         """
         self.configure_contours(levels, smoothing_mode, smoothing_factor)
         self.set_contour_dash(dash_mode, thickness)
         if color is not None:
-            self.call_action("contourConfig.setColor", color)
+            self.set_contour_color(color)
         if colormap is not None:
-            self.call_action("contourConfig.setColormap", colormap)
-            self.call_action("contourConfig.setColormapEnabled", True)
-            if bias is not None:
-                self.call_action("contourConfig.setColormapBias", bias)
-            if contrast is not None:
-                self.call_action("contourConfig.setColormapContrast", contrast)
-        else:
-            self.call_action("contourConfig.setColormapEnabled", False)
+            self.set_contour_colormap(colormap, bias, contrast)
         self.apply_contours()
 
     def clear_contours(self):
--- a/carta/image.py
+++ b/carta/image.py
@@ -116,7 +116,7 @@ class Image:
     def get_value(self, path):
         """Convenience wrapper for the session object's generic method for retrieving attribute values.
 
-        This method calls :obj:`carta.session.Session.get_value` after prepending this image's base path to the path parameter.
+        This method calls :obj:`carta.session.Session.get_value` after prepending this image's base path to the *path* parameter.
 
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


kswang1029 (approved):
this is great, handy, and intuitive. 👍

confluence:
I've realised that I haven't been consistent in mentioning default values in the function docstring (this was carried over from the original individual functions). I'd like to fix this before merging.

What are the units for the smoothing factor -- pixels? Does it depend on the smoothing mode?

kswang1029:
> I've realised that I haven't been consistent in mentioning default values in the function docstring (this was carried over from the original individual functions). I'd like to fix this before merging.
> 
> What are the units for the smoothing factor -- pixels? Does it depend on the smoothing mode?

It is in pixel as the kernel size. It is for Gaussian and box smoothing.

kswang1029:
> > I've realised that I haven't been consistent in mentioning default values in the function docstring (this was carried over from the original individual functions). I'd like to fix this before merging.
> > What are the units for the smoothing factor -- pixels? Does it depend on the smoothing mode?
> 
> It is in pixel as the kernel size. It is for Gaussian and box smoothing.


> > I've realised that I haven't been consistent in mentioning default values in the function docstring (this was carried over from the original individual functions). I'd like to fix this before merging.
> > What are the units for the smoothing factor -- pixels? Does it depend on the smoothing mode?
> 
> It is in pixel as the kernel size. It is for Gaussian and box smoothing.

Uh just realized that IIRC some parameters can be customized in the preferences so we need to consider this in the docstring🤔

confluence:
Ah, if the default is configurable in the preferences then I think I should rewrite these functions not to hardcode a default. There are probably a few other functions like that.

kswang1029:
> Ah, if the default is configurable in the preferences then I think I should rewrite these functions not to hardcode a default. There are probably a few other functions like that.

Guess we need a preferences class to store all user defaults then we can refer it elsewhere in the codebase?

confluence:
> Guess we need a preferences class to store all user defaults then we can refer it elsewhere in the codebase?

It should be possible to write these functions in such a way that the frontend's default is used if the user doesn't specify a value, and the wrapper doesn't need to be aware of the default. I think that in some places I copied a default from the frontend, assuming that it was hardcoded there, but even for defaults that are hardcoded we should avoid doing this (because then they can get out of sync, and have to be maintained in two places). There are some exceptions which are unlikely to change and are probably fine.

I'm pretty sure that in cases where the same action is used to set multiple values (and the function calling that action needs to supply all the parameters), we can use the existing macro functionality in the wrapper to refer to values set in the frontend without ever specifying the literal values. I discussed this with @I-Chenn yesterday because it came up in the vector overlay PR; we still need to check the solution to make 100% sure that it works. I'll have a closer look at this on Monday.

confluence:
@kswang1029 I have modified the functions so that all of `levels`, `smoothing_mode` and `smoothing_factor` are optional. This means that we avoid hardcoding defaults which can be changed in the frontend, and also that we can now specify only one or two of the three both in `configure_contours` and in `plot_contours`. The others default to the current frontend value. This is more consistent with the way that the style properties work.

I have noticed a discrepancy in the behaviour of various parts of this configuration in the UI. The style properties (dash and colour) are updated in the displayed image immediately when their respective actions are called, *and* the changed values are visible in the GUI contour dialog. However, the levels and smoothing options are only updated in the image when the action to apply contours is called (something which we can mitigate in the wrapper) *and* the changed values are *never* visible in the GUI dialog (which may be confusing to the user). I don't know how much of a problem the latter is. This is frontend behaviour (those sections of the dialog do not seem to detect when the underlying values have changed in the same way as the style section does), so it would need to be addressed in the frontend.

I am considering whether we should always call the `applyContours` action from the `configure_contours` method and remove the separate `apply_contours` method. Now that we have `plot_contours`, perhaps it's fine to keep the more fine-grained methods as they are -- but this makes the behaviour of the levels and smoothing options inconsistent with the behaviour of the style options. Do you have a preference?

kswang1029:
> @kswang1029 I have modified the functions so that all of `levels`, `smoothing_mode` and `smoothing_factor` are optional. This means that we avoid hardcoding defaults which can be changed in the frontend, and also that we can now specify only one or two of the three both in `configure_contours` and in `plot_contours`. The others default to the current frontend value. This is more consistent with the way that the style properties work.
> 
> I have noticed a discrepancy in the behaviour of various parts of this configuration in the UI. The style properties (dash and colour) are updated in the displayed image immediately when their respective actions are called, _and_ the changed values are visible in the GUI contour dialog. However, the levels and smoothing options are only updated in the image when the action to apply contours is called (something which we can mitigate in the wrapper) _and_ the changed values are _never_ visible in the GUI dialog (which may be confusing to the user). I don't know how much of a problem the latter is. This is frontend behaviour (those sections of the dialog do not seem to detect when the underlying values have changed in the same way as the style section does), so it would need to be addressed in the frontend.
> 
> I am considering whether we should always call the `applyContours` action from the `configure_contours` method and remove the separate `apply_contours` method. Now that we have `plot_contours`, perhaps it's fine to keep the more fine-grained methods as they are -- but this makes the behaviour of the levels and smoothing options inconsistent with the behaviour of the style options. Do you have a preference?

I feel having high level functions is sufficient based on the general use cases. The inconsistency in “apply” is due to that the responsible end is different. Changing contour color/thickness is using WebGL so it is pure frontend and it is sensible to see the results immediately (GPU does the job very fast). Smoothing method however involves the backend for image smoothing before calculating contour vertices. I think it’s okay to have such “inconsistencies” even in the python interface. Mostly I think users will use high level function directly.

confluence:
OK, great. In that case, I'm going to merge this -- I tested the optional parameters at my end and they seem to be working as expected.

PR #83, opened 2023-03-13
Icchen/79 render vector overlay

### LINKED ISSUE #79: support vector overlay rendering
Similar to the contour rendering, the Image class needs a set of low-level functions for configuration and a high-level wrapper of low-level functions.

I-Chenn:
This PR is to resolve #79.

@kswang1029  @confluence, I haven't finished this issue yet, but would like to ask a question before going on.

I have added `configure_vector_overlay`, `set_vector_overlay_color`, `set_vector_overlay_colormap`, `apply_vector_overlay` and `clear_vector_overlay` to `image.py`.

The `configure_vector_overlay` is a huge method with 9 parameters within since `VectorOverlayStore` in the frontend stuck all the parameters into `setVectorOverlayConfiguration` action to set the `Configuration` panel. Nevertheless, for the `Styling` panel, it seems that all the parameters are divided into respective actions.

My question is: do you prefer the way of dividing these actions into different Python functions? or do you prefer to stuck all these actions into one function , like a `set_vector_overlay_styling` function, for setting up the styling panel?


confluence:
I think that this split is OK for now. You can, however, eliminate the boolean parameters enabling or disabling certain features, because you can deduce them from the other parameters. Set `pixel_averaging_enabled` to `True` if `pixel_averaging` is not `None`, otherwise `False`, and so on (in this particular case, make sure you change the validation for `pixel_averaging` so that `None` is allowed). You can do the same for `threshold`.

For debiasing, are both error values required? If the user specifies one, does it make sense to use it for both errors, or are they likely to be completely different magnitudes? Does either of them have a sensible default? Depending on the answer to this, you can enable debiasing either if both error parameters are given, or if at least one is given.

Everything under Styling (except the colour, which I think it makes sense to keep separate for consistency) should maybe be grouped under a `set_vector_overlay_line`, or maybe `set_vector_overlay_style` (in which case `set_contour_dash` should perhaps be renamed to `set_contour_style` to match).

kswang1029:
> I think that this split is OK for now. You can, however, eliminate the boolean parameters enabling or disabling certain features, because you can deduce them from the other parameters. Set `pixel_averaging_enabled` to `True` if `pixel_averaging` is not `None`, otherwise `False`, and so on (in this particular case, make sure you change the validation for `pixel_averaging` so that `None` is allowed). You can do the same for `threshold`.
> 
> For debiasing, are both error values required? If the user specifies one, does it make sense to use it for both errors, or are they likely to be completely different magnitudes? Does either of them have a sensible default? Depending on the answer to this, you can enable debiasing either if both error parameters are given, or if at least one is given.

errors for stokes Q and U could be different so we need them as separate parameters.


> 
> Everything under Styling (except the colour, which I think it makes sense to keep separate for consistency) should maybe be grouped under a `set_vector_overlay_line`, or maybe `set_vector_overlay_style` (in which case `set_contour_dash` should perhaps be renamed to `set_contour_style` to match).

Vector overlay can be line segments or filled squares. Need to consider this in the naming convention.

confluence:
> errors for stokes Q and U could be different so we need them as separate parameters.

Yes, there will definitely be two parameters. But is the user ever likely to use the same value for both, or is it very unlikely?

(Actually, it probably doesn't matter -- it would be way too magical to make one default to the other, so we shouldn't do it either way. :) )

kswang1029:
> > errors for stokes Q and U could be different so we need them as separate parameters.
> 
> Yes, there will definitely be two parameters. But is the user ever likely to use the same value for both, or is it very unlikely?
> 
> (Actually, it probably doesn't matter -- it would be way too magical to make one default to the other, so we shouldn't do it either way. :) )

If stokes Q and U images are created with different clean processes (clean regions, depth, etc), errors can be different. I think we can set them as 0 as default and they are only effective when debiasing is True.

confluence:
We're trying to eliminate the boolean parameters and deduce them from the options. So in this case, we need to decide if we need *both* of the errors to be provided, or just one. Will it ever make sense to use one error value and one default error of zero, or are both errors always required to do anything meaningful?

kswang1029:
> We're trying to eliminate the boolean parameters and deduce them from the options. So in this case, we need to decide if we need _both_ of the errors to be provided, or just one. Will it ever make sense to use one error value and one default error of zero, or are both errors always required to do anything meaningful?

We need both to be provided if debiasing is True. If one is missing, we should throw an error.

confluence:
OK, but we don't want the user to have to duplicate information in the high-level API by enabling debiasing explicitly -- we want to deduce whether the user wants to enable debiasing from the debiasing options provided. So if both errors are always needed, it's fine to keep the defaults of `None`, and enable debiasing if *both* the error parameters are provided. This is in line with removing e.g. the flag that enables the threshold, and enabling the threshold automatically if a threshold value is provided.

If the user provides only one of the errors, we should disable debiasing, but also print a warning (since this is likely to be a mistake).

I-Chenn:
> I think that this split is OK for now. You can, however, eliminate the boolean parameters enabling or disabling certain features, because you can deduce them from the other parameters. Set `pixel_averaging_enabled` to `True` if `pixel_averaging` is not `None`, otherwise `False`, and so on (in this particular case, make sure you change the validation for `pixel_averaging` so that `None` is allowed). You can do the same for `threshold`.

@confluence, sorry for the late reply since I am a bit busy this evening. I will take a close look for all your comments tomorrow but would like to make a quick confirmation:

Is the modification in 32aae92 what you mean in the quotation above?

confluence:
> Is the modification in [32aae92](https://github.com/CARTAvis/carta-python/commit/32aae923ac3a5c85a4d0221e7ebde70de710c50f) what you mean in the quotation above?

Yes, that's correct! I also suggest that you do the same for `debiasing`, but as discussed, it should be enabled if both the error parameters are not `None`. If only one error parameter is provided, the debiasing should be disabled, but a warning should be printed.

I'll also make a few review comments.

carta/image.py
```diff
@@ -593,6 +593,87 @@ def hide_contours(self):
...
+            To enable or disable debiasing.
+        qError : {6}
+            The stoke Q Error.
+        uError : {7}
+            The stoke U Error.
+        """
+        if pixel_averaging is not None:
+            pixel_averaging_enabled = True
+        else:
+            pixel_averaging_enabled = False
```
confluence: You can simplify this and the conditional below like this: `pixel_averaging_enabled = pixel_averaging is not None`.

carta/image.py
```diff
@@ -593,6 +593,87 @@ def hide_contours(self):
         """Hide the contours."""
         self.set_contours_visible(False)
 
+    # VECTOR OVERLAY
+
+    @validate(Constant(VectorOverlaySource), Constant(VectorOverlaySource), NoneOr(Number()), Boolean(), NoneOr(Number()), Boolean(), NoneOr(Number()), NoneOr(Number()))
+    def configure_vector_overlay(self, angular_source=VectorOverlaySource.CURRENT, intensity_source=VectorOverlaySource.CURRENT, pixel_averaging=4, fractional_intensity=False, threshold=None, debiasing=False, qError=None, uError=None):
```
confluence: Naming convention: `qError` and `uError` should be `q_error` and `u_error`.

carta/image.py
```diff
@@ -593,6 +593,87 @@ def hide_contours(self):
...
+        """Configure vector overlay.
+
+        Parameters
+        ----------
+        angular_source : {0}
+            The angular source.
+        intensity_source : {1}
+            The intensity source.
+        pixel_averaging : {2}
+            The pixel averaging width. Set it to None to disable the usage pixel averaging width.
```
confluence: When you mention literals like `None` in the code, use the ReST markup for code, which is double backticks:

```
blah blah ``None`` ...
```

This will format the word as code in the final Sphinx document.

confluence: This should also mention that the default is `4` (plus the units; not sure what those are).

carta/image.py
```diff
@@ -593,6 +593,87 @@ def hide_contours(self):
...
+        fractional_intensity : {3}
+            Set polarization intensity to absolute or fractional.
```
confluence: I'd rephrase this as: "Enable fractional polarization intensity. By default absolute polarization intensity is used."

carta/image.py
confluence: I would change the last sentence to: "Set to None to disable pixel averaging."

carta/image.py
```diff
@@ -593,6 +593,87 @@ def hide_contours(self):
...
+    @validate(Constant(VectorOverlaySource), Constant(VectorOverlaySource), NoneOr(Number()), Boolean(), NoneOr(Number()), Boolean(), NoneOr(Number()), NoneOr(Number()))
+    def configure_vector_overlay(self, angular_source=VectorOverlaySource.CURRENT, intensity_source=VectorOverlaySource.CURRENT, pixel_averaging=4, fractional_intensity=False, threshold=None, debiasing=False, qError=None, uError=None):
+        """Configure vector overlay.
+
+        Parameters
+        ----------
+        angular_source : {0}
+            The angular source.
+        intensity_source : {1}
+            The intensity source.
```
confluence: Both of these should mention that the default is the current image.

carta/image.py
```diff
@@ -593,6 +593,87 @@ def hide_contours(self):
...
+        pixel_averaging : {2}
+            The pixel averaging width. Set it to None to disable the usage pixel averaging width.
+        fractional_intensity : {3}
+            Set polarization intensity to absolute or fractional.
+        threshold : {4}
+            The threshold. Set it to None to disable the usage of threshold.
```
confluence: I would say: The threshold. By default the threshold is disabled.

(We should probably also mention the units here.)

(If the user doesn't want to enable the threshold, they are more likely to omit the parameter and fall back to the default than to pass in an explicit `None`.)

carta/image.py
```diff
@@ -593,6 +593,87 @@ def hide_contours(self):
...
+        debiasing : {5}
+            To enable or disable debiasing.
+        qError : {6}
+            The stoke Q Error.
+        uError : {7}
+            The stoke U Error.
```
confluence: stoke -> Stokes, but Error -> error

Once the `debiasing` parameter is removed, I would revise each of these to be something like: "The Stokes _ error in [units]. Set both this and [other variable name] to enable debiasing. By default debiasing is disabled."

(Edited to swap sentence order in suggested wording and simplify it.)

confluence (commented):
Some nitpicks about the phrasing in the docstring.

I-Chenn:
@kswang1029 @confluence 
I have made a commit (cb56287) to make several modifications to `configure_vector_overlay` function following your comments. Nevertheless, I found some further problems and would like to discuss with you:
1. `angular_source` and `intensity_source` setting:
These two parameters actually do not set "Current" as default. It depends on what kind of image you open; if an image with polarization info is opened, these two parameters are set to "Computed" as default automatically. See below:
image
May I have some hint of setting up the default values of `angular_source` and `intensity_source` under such circumstance?

2. The management of the situation "`q_error` is set but `u_error` is not set" (or vise versa):
I am now dealing with the situation with the code:
image
But do I need to set `debiasing` to some `None` value rather than just print the message?

3. `Intensity` area in the `Styling` panel:
These two input fields seem to automatically update themselves after we apply the vector overlay. I don't know if we need to fill them up manually in advance (perhaps for some circumstances) or we just keep it a `None`?
image

carta/image.py
```diff
@@ -593,6 +593,83 @@ def hide_contours(self):
...
+            The Stokes U error in Jy/beam. Set both this and ``q_error`` to enable debiasing. The debiasing is disabled by default.
+        """
+        pixel_averaging_enabled = pixel_averaging is not None
+        threshold_enabled = threshold is not None
+        if q_error is not None and u_error is not None:
+            debiasing = True
+        elif q_error is None and u_error is None:
+            debiasing = False
+        else:
+            print("q_error and u_error must be both set to enable debiasing.")
```
confluence: Please use the logger for this message (you'll have to import it at the top; it's in `.util` -- see how it's used in e.g. `session.py`).

You could compact this using syntax similar to the previous lines (as shown below), but it does mean restating part of the same conditional, so I don't feel that strongly about it -- maybe the way it is now is clearer to read.

```
debiasing = q_error is not None and u_error is not None
if not debiasing and (q_error is not None or u_error is not None):
    logger.warning(...)
```

In the message, please swap "be both" -> "both be". I would also refer to the error variables by their descriptions rather than their variable names here ("The Stokes Q error and the Stokes U error must...")

carta/image.py
```diff
@@ -593,6 +593,83 @@ def hide_contours(self):
...
+        pixel_averaging : {2}
+            The pixel averaging width in pixel. Set to ``None`` to disable pixel averaging.
+        fractional_intensity : {3}
+            Enable fractional polarization intensity. By default the absolute polarization intensity is used.
+        threshold : {4}
+            The threshold in Jy/pixel. By default the threshold is disabled.
+        q_error : {5}
+            The Stokes Q error in Jy/beam. Set both this and ``u_error`` to enable debiasing. The debiasing is disabled by default.
+        u_error : {6}
+            The Stokes U error in Jy/beam. Set both this and ``q_error`` to enable debiasing. The debiasing is disabled by default.
```
confluence: "The debiasing..." should just be "Debiasing..." in the last sentence(s) here.

carta/image.py
```diff
@@ -593,6 +593,83 @@ def hide_contours(self):
...
+    def configure_vector_overlay(self, angular_source=VectorOverlaySource.CURRENT, intensity_source=VectorOverlaySource.CURRENT, pixel_averaging=4, fractional_intensity=False, threshold=None, q_error=None, u_error=None):
+        """Configure vector overlay.
+
+        Parameters
+        ----------
+        angular_source : {0}
+            The angular source. By default the current image is used.
+        intensity_source : {1}
+            The intensity source. By default the current image is used.
+        pixel_averaging : {2}
```
confluence: "in pixel" -> "in pixels"

confluence: (And it needs the default -- I think it's fine to put "The default is 4." as the middle sentence; you don't need to restate the units.)

confluence (commented):
Minor followup comments; it's looking good!

carta/image.py
confluence: Sorry; I selected the wrong lines somehow -- this is related to the pixel averaging description.

confluence:
I made my previous comments before I saw your questions -- answers below.

1. Perhaps the default parameter should be `None`, ~and the default source determined in the function using the same criteria that the frontend uses. You can use the image functions for determining if particular polarizations are available.~ The way that this default is selected would then be described in the documentation for the parameter. Edit: I was overthinking this. You should be able to get the default that was already computed by the frontend using `get_value`.

2. The debiasing can't be undefined -- it's either on or off (and it should be off unless both errors are provided). I suggested how you could rewrite this conditional -- in that syntax, debiasing is already set to `False` unless both errors are not `None`. If you keep the current syntax, you should also set it to `False` explicitly when you print the warning.

3. It looks like these intensity defaults are computed from the file. I don't think that this is necessarily going to be a problem. There's a single action for setting both the min and the max. We're planning to have a single style function for calling this and a few other actions (although maybe we should split this up further). If the user doesn't provide any intensity parameters we don't have to call this action at all; if they provide both parameters then we override both at the same time; if they provide only one, we can fetch the other one using `get_value` so that we don't override it. The only thing I'm not sure about is whether those defaults will be populated correctly if we apply the overlay from the scripting interface -- we should test that to make sure.

I-Chenn:
@confluence , I added `set_vector_overlay_style`, `plot_vector_overlay`, `set_vector_overlay_visible`, `show_vector_overlay`, `hide_vector_overlay` to `image.py` and made some modifications to `configure_vector_overlay` following your suggestions in 401652e.
I haven't dealt with the action that need to be taken if user set one of the input field for `Intensity` area of `set_vector_overlay_style` yet, but the function is workable with the default setting under my simple test.
Nevertheless, I also tested `configure_vector_overlay`, but there seems to be some error at the last line of that function (the `call_action` line). I haven't figured out what the problem is and will work on it tomorrow.
Please take a look at the parameter descriptions in `set_vector_overlay_style` and see if it's OK for you if it's convenient to you. Thank you very much.

carta/image.py
```diff
@@ -593,6 +593,181 @@ def hide_contours(self):
...
+            self.call_action("vectorOverlayConfig.setColor", color)
+        if colormap is not None:
+            self.call_action("vectorOverlayConfig.setColormap", colormap)
+            self.call_action("vectorOverlayConfig.setColormapEnabled", True)
+            if bias is not None:
+                self.call_action("vectorOverlayConfig.setColormapBias", bias)
+            if contrast is not None:
+                self.call_action("vectorOverlayConfig.setColormapContrast", contrast)
+        else:
+            self.call_action("vectorOverlayConfig.setColormapEnabled", False)
```
confluence: I realised that I made this unnecessarily complicated in my `plot_contours` PR, and have just updated it to reflect this. I was avoiding disabling the colormap in `set_color` -- but it doesn't actually matter, as long as you call `set_color` first and then `set_colormap` (because if they are both present, the colormap is shown). So you can do the same here, and first call `set_color` (if `color` is not `None`) and then `set_colormap` (if `colormap` is not `None`).

carta/image.py
```diff
@@ -593,6 +593,181 @@ def hide_contours(self):
...
+            The threshold in Jy/pixels. By default is 4.
+        q_error : {5}
+            The Stokes Q error in Jy/beam. Set both this and ``u_error`` to enable debiasing. Debiasing is disabled by default.
+        u_error : {6}
+            The Stokes U error in Jy/beam. Set both this and ``q_error`` to enable debiasing. Debiasing is disabled by default.
+            thickness : {0}
+            The line thickness in pixels. By default is 1.
+        intensity_min : {1}
+            The minimum value of intensity in Jy/pixel.
+        intensity_min : {2}
```
confluence: This numbering was also a bug in my contour PR -- I just fixed it. You need to renumber these placeholders to match the order of the parameters in this method.

`thickness` also needs to be de-indented by one level.

I-Chenn: Modification is made in 26aac1f.

carta/image.py
I-Chenn: Modification is made in 26aac1f.

I-Chenn:
@confluence @kswang1029, I bumped into some problems and would like to ask some questions about them
The main question is:
**Is it possible to serialize the `None` at Python side to the `undefined` at TS/JS side when using `call_action`?**

As I discussed with @confluence few days ago, I made some tests by executing:
```
print(img.get_value("vectorOverlayConfig.intensityMin"))
print(img.get_value("vectorOverlayConfig.intensityMax"))

img.call_action("vectorOverlayConfig.setIntensity", None, None)

print(img.get_value("vectorOverlayConfig.intensityMin"))
print(img.get_value("vectorOverlayConfig.intensityMax"))
```
in my test script, and I got:
```
None
None
  File "/home/icchen/Python_wrappers/test.py", line 19, in <module>
    img.call_action("vectorOverlayConfig.setIntensity", None, None)
  File "/home/icchen/carta-build/venv_carta-python/lib/python3.10/site-packages/carta/image.py", line 114, in call_action
    return self.session.call_action(f"{self._base_path}.{path}", *args, **kwargs)
  File "/home/icchen/carta-build/venv_carta-python/lib/python3.10/site-packages/carta/session.py", line 250, in call_action
    return self._protocol.request_scripting_action(self.session_id, path, *args, **kwargs)
  File "/home/icchen/carta-build/venv_carta-python/lib/python3.10/site-packages/carta/protocol.py", line 372, in request_scripting_action
    raise CartaActionFailed(f"{carta_action_description} failed: {response_data.get('message', 'No error message received.')}")
carta.util.CartaActionFailed: CARTA scripting action frameMap[17].vectorOverlayConfig.setIntensity called with parameters (None, None) failed: No error message received.
```
The result shows that I can deserialize the `undefined` at TS/JS side to the `None` at Python side when using `get_value`, but fail to serialize the `None` at Python side to the `undefined` at TS/JS side with `call_action`.

The reason I am asking this question is that I must fill in the default values (including `None`)  for all the arguments when dealing with `configure_vector_overlay` function. I did some survey in the `image.py` for how to deal with `None` as default when using `call_action`, and in most cases, it is like:
```
if [parameter] is not None:
    self.call_action([parameter_related_action], parameter)
```
to get rid of the direct usage of `None` in `call_action`. 
Nevertheless, since the `setVectorOverlayConfiguration` at frontend is a huge function that sets everything at once, it's impossible to deal with the "default `None`" situation with the method mentioned above.

Analogous situation appear when calling the `setIntensityRange` in `set_vector_overlay_style` function. I can now deal with the default values of the parameters when user "set both the parameters" or "skip filling both parameters", however, if user fill in one of the parameters, I still need to call `setIntensityRange` with the other parameter set to `None`.

I also made some tests with `Macro` function suggested by @confluence, but it doesn't seem to work.

Thus, if it's possible to serialize the `None` at Python side to the `undefined` at the TS/JS side, it would be easier to set `None` directly when calling frontend action with `call_action`.

confluence:
Observing this behaviour only through the wrapper is obscuring what is actually happening in the various serialization and deserialization steps. You need to inspect the scripting request and response messages logged by the frontend.

We serialize to JSON (in both Python and JS) and deserialize from JSON (in both Python and JS). The JS `undefined` value is not directly serializable, so it is serialized to `null` in the JS. Python also serializes `None` to `null` -- so far, so good.

The `CartaActionFailed` exception is caused by a failure response in the frontend. In this particular case, if we look at the scripting response to a call of this action with one or more `None` parameters, we see the error `TypeError: e is null` -- so the null value is being rejected either by the scripting service or the action that it is calling (I haven't worked this out; I also need to check why this error message is not being passed to the wrapper correctly).

If we bypass the serialization by using a macro which evaluates to `undefined` (e.g. the value of one of the intensity properties before they are set), or apparently if we pass any *non-empty* string value (`"a"`, etc.), that does appear to unset the property and result in identical overlay settings (an *empty* string is evaluated as zero).

I have noticed that the GUI inputs for the intensity values do not update consistently when the overlay is not applied (in ways which are not related to scripting -- for example, if you enter a value while the overlay is not applied and then click on the button to clear it, it is not cleared). In previous tests this made it difficult to see if the values were being set and unset correctly, and I thought that there were circumstances under which they were set and then ignored. However, I can't currently reproduce a situation where the values are set through the wrapper and *not* applied correctly.

This does leave the question of what the correct thing to do is. Do we want a `None` value passed in for one of the values to default to the automatic value or default to the current value? I would say that the first option makes more sense -- which means that we should explicitly send a value to unset the property. Currently it seems that we could use an arbitrary (non-empty!) string for this, but I'm concerned that this is a side effect of using the wrong variable type, and we should not rely on it working like this. I would like to find out exactly why the null value is being rejected (whether it is happening in the scripting service or the function), and find a way to send a value which will explicitly evaluate to `undefined`.

I-Chenn:
@confluence, @kswang1029 
Following @confluence's suggestion, I made some modifications to deal with `None` (for now):
Still setting `None` for the function arguments, but then convert it into "non-empty string" (I adopt "None") in the function and use this converted string to process the `call_action`.
I did some tests and was able to run `configure_vector_overlay`, `set_vector_overlay_style` and the high level function `plot_vector_overlay` with their default settings, like:
```
img = session.open_image("Orion_source_I_Stokes_IQU.fits")

img.configure_vector_overlay()
img.set_vector_overlay_style()
img.plot_vector_overlay()
```
Please let me know if you like it or not.

carta/image.py
```diff
@@ -605,6 +605,189 @@ def hide_contours(self):
...
+        if angular_source is None:
+            angular_source = "None"
+        if intensity_source is None:
+            intensity_source = "None"
+        if threshold is None:
+            threshold = "None"
+        if q_error is None:
+            q_error = "None"
+        if u_error is None:
+            u_error = "None"
```
confluence: All of these `"None"`s need to be replaced by macros -- this hack should *only* be used when it's necessary to explicitly *unset* a property (which is only the case with the intensity range values). In this case we *don't* want to unset these properties if the parameters are `None`; we want to keep the properties which are currently set in the frontend.

confluence (changes requested):

confluence:
I realise now that I may have given you conflicting suggestions about handling the defaults. If we have e.g. a single `threshold` parameter, which disables the threshold if it's `None` and enables it if it's not `None`, then we can't also let the user omit the parameter (which sets it to `None`) to keep the current frontend setting (which may be enabled). The latter behaviour is consistent with the rest of the wrapper API, and I think we should keep it, but I also don't want the user to have to set redundant parameters. So I have an updated suggestion which I think is a compromise between these requirements.

1. Put back the explicit flags for enabling pixel averaging, threshold and debiasing.
2. Set all the parameter defaults to `None` (including the pixel averaging and fractional intensity -- we shouldn't be hardcoding and setting those defaults here).
3. This is my updated suggestion for the parameter logic in `configure_vector_overlay`:

```
# Set flags to True only if options are not None and flags are None
if pixel_averaging is not None and pixel_averaging_enabled is None:
    pixel_averaging_enabled = True
if threshold is not None and threshold_enabled is None:
    threshold_enabled = True
if q_error is not None and u_error is not None and debiasing is None:
    debiasing = True

# Disable debiasing and warn on missing error
if (q_error is not None and u_error is None) or (q_error is None and u_error is not None):
    debiasing = False
    logger.warning("The Stokes Q error and Stokes U error must both be set to enable debiasing.")
    
# Now fall back to the current frontend values for anything that is still None
if angular_source is None:
    angular_source = self.macro("vectorOverlayConfig", "angularSource")
if intensity_source is None:
    intensity_source = self.macro("vectorOverlayConfig", "intensitySource")
if pixel_averaging_enabled is None:
    pixel_averaging_enabled = self.macro("vectorOverlayConfig", "pixelAveragingEnabled")
if pixel_averaging is None:
    pixel_averaging = self.macro("vectorOverlayConfig", "pixelAveraging")
if fractional_intensity is None:
    fractional_intensity = self.macro("vectorOverlayConfig", "fractionalIntensity")
if threshold_enabled is None:
    threshold_enabled = self.macro("vectorOverlayConfig", "thresholdEnabled")
if threshold is None:
    threshold = self.macro("vectorOverlayConfig", "threshold")
if debiasing is None:
    debiasing = self.macro("vectorOverlayConfig", "debiasing")
if q_error is None:
    q_error = self.macro("vectorOverlayConfig", "qError")
if u_error is None:
    u_error = self.macro("vectorOverlayConfig", "uError")
    
# then call the action...
```

So by default, if a parameter or all related parameters are `None`, we use macros to keep whatever is currently set in the frontend. The errors may be unset, but this will not trigger the crash because we're using a macro to resolve the value in the frontend.

E.g. if the user provides a value for `threshold`, we deduce that `threshold_enabled` should be set to `True`, *unless* the user has explicitly set it to `False` (an explicit flag should take precedence over the value). This is similar to our assumptions elsewhere that if the user provides a colourmap they want to use the colourmap, and that if they provide a colour they do not want to use the colourmap.

We enable debiasing if both errors are provided and debiasing is not explicitly disabled. If only one error is provided, we disable debiasing and print a warning.

The docstring should be updated accordingly -- I think that we should keep the explanation of the defaults for the sources, but reword it to something like: `If this is unset, the default is computed PA if the image contains Stokes information, or the current image otherwise.`. But we can omit the defaults for most of the other values except the ones which have special logic, e.g.:

* `pixel_averaging_enabled`: (...) By default this is enabled if `pixel_averaging` is set.
* `pixel_averaging`: (...) If this is set, pixel averaging is enabled by default.

I'll add a few comments about the other functions.

carta/image.py
```diff
@@ -605,6 +605,189 @@ def hide_contours(self):
...
+        self.call_action("vectorOverlayConfig.setVectorOverlayConfiguration", angular_source, intensity_source, pixel_averaging_enabled, pixel_averaging, fractional_intensity, threshold_enabled, threshold, debiasing, q_error, u_error)
+
+    @validate(Number(), NoneOr(Number()), NoneOr(Number()), Number(), Number(), Number())
+    def set_vector_overlay_style(self, thickness=1, intensity_min=None, intensity_max=None, length_min=0, length_max=20, rotation_offset=0):
```
confluence: For consistency with the approach of ignoring options which are not explicitly provided (and keeping the currently set frontend values), all these hardcoded defaults should be removed, and all these parameters should default to `None`.

You don't need to use any macros in here, but you do need to add guards to all the individual action calls, and only call the actions if the parameters are not `None`. (The intensity range logic is fine; the action won't be called if both values are `None`, which is correct.)

carta/image.py
```diff
@@ -605,6 +605,189 @@ def hide_contours(self):
...
+            self.call_action("vectorOverlayConfig.setColormapBias", bias)
+        if contrast is not None:
+            self.call_action("vectorOverlayConfig.setColormapContrast", contrast)
+
+    def apply_vector_overlay(self):
+        """Apply the vector overlay configuration."""
+        self.call_action("applyVectorOverlay")
+
+    @validate(NoneOr(Constant(VectorOverlaySource)), NoneOr(Constant(VectorOverlaySource)), NoneOr(Number()), Boolean(), NoneOr(Number()), NoneOr(Number()), NoneOr(Number()), Number(), NoneOr(Number()), NoneOr(Number()), Number(), Number(), Number(), NoneOr(Color()), NoneOr(Constant(Colormap)), NoneOr(Number()), NoneOr(Number()))
+    def plot_vector_overlay(self, angular_source=None, intensity_source=None, pixel_averaging=4, fractional_intensity=False, threshold=None, q_error=None, u_error=None, thickness=1, intensity_min=None, intensity_max=None, length_min=0, length_max=20, rotation_offset=0, color=None, colormap=None, bias=None, contrast=None):
```
confluence: In this function the parameters and docstring need to be updated and all the parameters also need to default to `None`, but the logic still looks correct.

confluence (changes requested):

carta/image.py
confluence: Hmm. Actually, there's still a problem here with the intensity. There needs to be a way for the user to explicitly unset *both* values (so that they are *both* automatically calculated), and currently if both parameters are omitted or explicitly set to `None`, the function won't be called at all, so the current settings will be kept. Also, if one intensity parameter is given and the other is not, we should be keeping the existing parameter from the frontend, not unsetting it explicitly, for consistency with the other parameters and methods.

I think that we need to permit an additional "auto" value for each intensity parameter. The cleanest way to implement this is to create a single-value enum in `constants.py` which inherits from `str`:

```
class Auto(str, Enum):
    """Special value for parameters to be calculated automatically."""
    AUTO = "Auto"
```

I made it generic because we may want to reuse it in other methods (for example, we could eliminate the `clear_color` session method for overlay components, and add `Auto.AUTO` as an allowed colour value in `set_color` instead -- but that's outside the scope of this PR). We could also use a plain string, but that's less user-friendly and more error-prone.

Then these parameters could both be validated by: `NoneOr(OneOf(Number(), Constant(Auto)))`. The logic in the method would then be:

```
if intensity_min is not None and intensity_max is not None:
    self.call_action("vectorOverlayConfig.setIntensityRange", intensity_min, intensity_max)
elif intensity_min is None and intensity_max is not None:
    self.call_action("vectorOverlayConfig.setIntensityRange", self.macro("vectorOverlayConfig.intensityMin"), intensity_max)
elif intensity_min is not None and intensity_max is None:
    self.call_action("vectorOverlayConfig.setIntensityRange", intensity_min, self.macro("vectorOverlayConfig.intensityMax"))
```

Now if one value is provided the other is kept as-is (if the existing value is unset, this should not trigger the crash because the macro is evaluated on the frontend). If either value is `Auto.AUTO`, the first part of the conditional will trigger, and at the moment you don't have to do anything else, because this object will be serialized as its string value `"Auto"`, which will have the same effect as `"None"`. Later if we can figure out what the problem is with sending `None` values, we'll be able to rewrite this to handle `Auto.AUTO` differently, but the user's API won't have to change.

In the parameter description we'd put e.g.
```
intensity_min: (...) Use :obj:`carta.constants.Auto.AUTO` to clear the custom value and calculate it automatically.
```

confluence (changes requested):

I-Chenn:
@confluence 
In my last few commits, I can successfully setting `configure_vector_overlay` with the `Macro` generated default values.
Nevertheless, there are still some errors for the `set_vector_overlay_style` with the use of `Macro`, I am still looking for the reason that causes the error.

I-Chenn:
@confluence 
I was able to let `set_vector_overlay_style` executed with its default setting in a250939. Nevertheless there are still some errors:

1. When I set desired value to `intensity_min` or `intensity_max`, there is error message like:
```
  File "/Users/i-chennchen/Python_wrappers/test.py", line 16, in <module>
    img.set_vector_overlay_style(intensity_min=0.1, intensity_max=0.5)
  File "/Users/i-chennchen/carta-build/venv_carta-python/lib/python3.9/site-packages/carta/validation.py", line 629, in newfunc
    raise CartaValidationFailed(f"Invalid function parameter: {msg}")
carta.util.CartaValidationFailed: Invalid function parameter: 0.1 is not one of <carta.validation.Number object at 0x1034d2d60>, <carta.validation.Constant object at 0x1034d2df0> or None.
```

2. When I execute the high-level function `plot_vector_overlay` with all the default values set in `configure_vector_overlay` and `set_vector_overlay_style`, the automatically generated `intensity_min` and `intensity_max` are different from what I get through GUI. For example, if I run:
```
img.plot_vector_overlay()
```
The generated intensity range is like:
截圖 2023-04-04 下午12 27 22

but if I do it manually with GUI and with the same image and same default setting, I got something like this:
截圖 2023-04-04 下午12 29 16
Not quite sure what went wrong but I will keep survey about it.

By the way, since I will be off to the computer this afternoon, if you have any further suggestions or comments, please feel free to address and I will keep working on it this evening or tomorrow.

confluence:
> 1. When I set desired value to `intensity_min` or `intensity_max`, there is error message like

Sorry, that's my bad -- I confused the validation classes `OneOf` and `Union`. `OneOf` combines specific values; `Union` is the one that combines other parameter classes, and is the one you should be using for these parameters.

> 2. When I execute the high-level function plot_vector_overlay with all the default values set...

Does this happen *only* if you call `plot_vector_overlay`, or does the same thing also happen if you call `set_vector_overlay_style` with default values (followed by `apply_vector_overlay`)?

I suggest that you also build the docs locally and check that the parameter validation types are what you expect (it's easy to get an off-by-one error when the lists get this long, and in the docs you will see the automatically generated type descriptions filled in).

I-Chenn:
> Sorry, that's my bad -- I confused the validation classes `OneOf` and `Union`. `OneOf` combines specific values; `Union` is the one that combines other parameter classes, and is the one you should be using for these parameters.
I'm sorry I don't quite understand how to use `Union`. I import it from .validation and then used it like:
```
NoneOr(Union(Number(), Constant(Auto)))
```
and then execute `set_vector_overlay_style()`  with the script like:
```
img = session.open_image("/Users/i-chennchen/fits/cluster_00512.fits")
img.set_vector_overlay_style()
```

but get an error message like:
```
  File "/Users/i-chennchen/Python_wrappers/test.py", line 1, in <module>
    from carta. Session import Session
  File "/Users/i-chennchen/carta-build/venv_carta-python/lib/python3.9/site-packages/carta/session.py", line 11, in <module>
    from .image import Image
  File "/Users/i-chennchen/carta-build/venv_carta-python/lib/python3.9/site-packages/carta/image.py", line 12, in <module>
    class Image:
  File "/Users/i-chennchen/carta-build/venv_carta-python/lib/python3.9/site-packages/carta/image.py", line 669, in Image
    def set_vector_overlay_style(self, thickness=None, intensity_min=None, intensity_max=None, length_min=None, length_max=None, rotation_offset=None):
  File "/Users/i-chennchen/carta-build/venv_carta-python/lib/python3.9/site-packages/carta/validation.py", line 644, in decorator
    newfunc.__doc__ = newfunc.__doc__.format(*(fix_description(p.description) for p in vargs))
  File "/Users/i-chennchen/carta-build/venv_carta-python/lib/python3.9/site-packages/carta/validation.py", line 644, in <genexpr>
    newfunc.__doc__ = newfunc.__doc__.format(*(fix_description(p.description) for p in vargs))
  File "/Users/i-chennchen/carta-build/venv_carta-python/lib/python3.9/site-packages/carta/validation.py", line 319, in description
    return self._description or " or ".join(o.description for o in self. Options)
TypeError: sequence item 0: expected str instance, Constant found
```
I then found with the use of `Union`, the above error occurs even when I just open image with the script:
```
img = session.open_image("/Users/i-chennchen/fits/cluster_00512.fits")
```
It seems to be a very bottom layer error. Did I right it right to import `Union` from validation? Or did I use it correctly?

> Does this happen _only_ if you call `plot_vector_overlay`, or does the same thing also happen if you call `set_vector_overlay_style` with default values (followed by `apply_vector_overlay`)?

If I call `set_vector_overlay_style` followed by `apply_vector_overlay`, I got exactly the same generated-value as the one I use GUI. Thus it seems to be the problem of `plot_vector_overlay`?

> I suggest that you also build the docs locally and check that the parameter validation types are what you expect (it's easy to get an off-by-one error when the lists get this long, and in the docs you will see the automatically generated type descriptions filled in).

It seems to be useful when dealing with long list of parameters, but I don't quite understand what you mean by "build the docs locally"?

confluence:
That error is happening in the function which is generating the description. This is because as the code is written now, the options have to be passed in as a single iterable parameter, not as multiple`*args` (so you need extra brackets). But I don't *like* that, and it's not consistent with `OneOf`, so I'm going to change it and commit a fix. You can add the extra brackets around `Number(), Constant(Auto)` now to get it to work, but I should have the change done soon (I just need to test it).

The docs on Read The Docs are built from the dev branch on every merge, but it's a good idea to preview them before you merge. You can do this in the `docs` directory with `make html` (after installing all the dependencies listed in the `requirements.txt` file there). You can then browse the built files in `build/html`.

confluence:
`NoneOr` actually inherits from `Union`, so while I'm in there I'll modify this so that you can just use `NoneOr(list, of, things, here)` instead of nesting a `Union` inside the `NoneOr`.

confluence:
Just merged in #99. You should now be able to do: `NoneOr(Number(), Constant(Auto))`
`

I-Chenn:
It works when setting `intensity_min` and `intensity_max` in`set_vector_overlay_style` with specified value in commit 91eac81. Thanks a lot!

confluence:
@I-Chenn once [#2173](https://github.com/CARTAvis/carta-frontend/pull/2173) is merged, it will be possible to fix the intensity range functionality:
* value is `None` -> use a macro to keep the current value
* value is `Auto.AUTO` -> use the special "undefined" macro to unset the value

We should create a subclass of `Macro` called `Undefined` which requires no parameters and constructs a macro with an `"undefined"` target parameter.

With this fixed, I think this PR will be ready for final review and merging.

carta/util.py
```diff
@@ -88,7 +88,14 @@ def __repr__(self):
     def json(self):
         """The JSON serialization of this object."""
         return {"macroTarget": self.target, "macroVariable": self.variable}
+    
+class Undefined(Macro):
+    """
+    A subclass of Macro to construct a placeholder for `"undefined"`.
+    """
 
+    def __init__(self, target="", variable="undefined"):
```
confluence: This constructor shouldn't take any parameters. You should just pass these literals into the parent constructor below.

I-Chenn:
@confluence, I have added a subclass `Undefined` of `Macro` in `util.py` and use it for `Auto.AUTO` cases in 285433b.

Nevertheless, I am not sure if I did it right to construct `Undefined`. When I use it right after opening an image, I can indeed set the intensity range to `Automatic` (or perhaps I just keep the default value?), nevertheless, when I did something like:
```
img.set_vector_overlay_style(intensity_min=1, intensity_max=2)
img.set_vector_overlay_style(intensity_min=Auto.AUTO, intensity_max=Auto.AUTO)
```
The `intensity_min` and `intensity_max` stay to the previous values (1 and 2 in this case) and I can't reset it to `Automatic`.

carta/image.py
```diff
@@ -672,6 +672,224 @@ def hide_contours(self):
...
+        elif intensity_min is None and intensity_max is not None:
+            self.call_action("vectorOverlayConfig.setIntensityRange", self.macro("vectorOverlayConfig", "intensityMin"), intensity_max)
+        elif intensity_min is not None and intensity_max is None:
+            self.call_action("vectorOverlayConfig.setIntensityRange", intensity_min, self.macro("vectorOverlayConfig", "intensityMax"))
+        elif intensity_min is Auto.AUTO and intensity_max is Auto.AUTO:
+            self.call_action("vectorOverlayConfig.setIntensityRange", Undefined(), Undefined())
+        elif intensity_min is not Auto.AUTO and intensity_max is Auto.AUTO:
+            self.call_action("vectorOverlayConfig.setIntensityRange", intensity_min, Undefined())
+        elif intensity_min is Auto.AUTO and intensity_max is not Auto.AUTO:
+            self.call_action("vectorOverlayConfig.setIntensityRange", Undefined(), intensity_max)
```
confluence: It's unnecessary to handle all these specific combinations (and you're currently not handling possible combinations of auto and None). You can simplify this logic a lot:

1. check if `intensityMin` is `None`, and if so set it to the macro to fetch the current value
2. otherwise check if it's `Auto.AUTO`, and if so set it to the undefined macro
3. do the same for `intensityMax`
4. then call the action with both (updated) values.

confluence (changes requested):

carta/image.py
```diff
@@ -672,6 +672,224 @@ def hide_contours(self):
...
+    # VECTOR OVERLAY
+
+    @validate(NoneOr(Constant(VectorOverlaySource)), NoneOr(Constant(VectorOverlaySource)), NoneOr(Boolean()), NoneOr(Number()), NoneOr(Boolean()), NoneOr(Boolean()), NoneOr(Number()), NoneOr(Boolean()), NoneOr(Number()), NoneOr(Number()))
+    def configure_vector_overlay(self, angular_source=None, intensity_source=None, pixel_averaging_enabled=None, pixel_averaging=None, fractional_intensity=None, threshold_enabled=None, threshold=None, debiasing=None, q_error=None, u_error=None):
+        """Configure vector overlay.
+
+        Parameters
+        ----------
+        angular_source : {0}
+            The angular source. If the image is with Stoke information, ``Computed PA`` is set by default; If the image is without Stoke information, ``Current image`` is set.
```
confluence: Please change this to:
```
The angular source. This is initially set to computed PA if the image contains Stokes information, otherwise to the current image.
```

carta/image.py
```diff
@@ -672,6 +672,224 @@ def hide_contours(self):
...
+        intensity_source : {1}
+            The intensity source. If the image is with Stoke information, ``Computed PI`` is set by default; If the image is without Stoke information, ``Current image`` is set.
```
confluence: Please change this description in the same way as the description above.

carta/image.py
```diff
@@ -672,6 +672,224 @@ def hide_contours(self):
...
+        pixel_averaging_enabled : {2}
+            To enable pixel averaging.
```
confluence: Change this to `Enable pixel averaging`.

carta/image.py
```diff
@@ -672,6 +672,224 @@ def hide_contours(self):
...
+        pixel_averaging : {3}
+            The pixel averaging width in pixel. Set to ``None`` to disable pixel averaging.
+        fractional_intensity : {4}
+            Enable fractional polarization intensity. By default the absolute polarization intensity is used.
+        threshold_enabled : {5}
+            To enable threshold.
```
confluence: Remove "to".

carta/image.py
```diff
@@ -672,6 +672,224 @@ def hide_contours(self):
...
+        threshold : {6}
+            The threshold in Jy/pixels. By default is 4.
+        debiasing : {7}
+            To enable debiasing.
```
confluence: Remove "to".

carta/image.py
```diff
@@ -672,6 +672,224 @@ def hide_contours(self):
...
+        pixel_averaging_enabled : {2}
+            To enable pixel averaging.
+        pixel_averaging : {3}
+            The pixel averaging width in pixel. Set to ``None`` to disable pixel averaging.
+        fractional_intensity : {4}
+            Enable fractional polarization intensity. By default the absolute polarization intensity is used.
+        threshold_enabled : {5}
+            To enable threshold.
+        threshold : {6}
+            The threshold in Jy/pixels. By default is 4.
```
confluence: Remove the second sentence.

carta/image.py
```diff
@@ -672,6 +672,224 @@ def hide_contours(self):
...
+        Parameters
+        ----------
+        angular_source : {0}
+            The angular source. If the image is with Stoke information, ``Computed PA`` is set by default; If the image is without Stoke information, ``Current image`` is set.
+        intensity_source : {1}
+            The intensity source. If the image is with Stoke information, ``Computed PI`` is set by default; If the image is without Stoke information, ``Current image`` is set.
+        pixel_averaging_enabled : {2}
+            To enable pixel averaging.
+        pixel_averaging : {3}
+            The pixel averaging width in pixel. Set to ``None`` to disable pixel averaging.
```
confluence: Remove the second sentence.

carta/image.py
```diff
@@ -672,6 +672,224 @@ def hide_contours(self):
         """Hide the contours."""
         self.set_contours_visible(False)
 
+    # VECTOR OVERLAY
+
+    @validate(NoneOr(Constant(VectorOverlaySource)), NoneOr(Constant(VectorOverlaySource)), NoneOr(Boolean()), NoneOr(Number()), NoneOr(Boolean()), NoneOr(Boolean()), NoneOr(Number()), NoneOr(Boolean()), NoneOr(Number()), NoneOr(Number()))
+    def configure_vector_overlay(self, angular_source=None, intensity_source=None, pixel_averaging_enabled=None, pixel_averaging=None, fractional_intensity=None, threshold_enabled=None, threshold=None, debiasing=None, q_error=None, u_error=None):
+        """Configure vector overlay.
+
```
confluence: In all these parameter descriptions we need to make a distinction between the *default value* -- which is always `None` and always results in the current frontend value being preserved, and the *initial value* set by the frontend, which is determined by the store constructor, and which may additionally be configurable in the frontend preferences.

This docstring needs to make sense whether someone is calling this function for the first time (when all the frontend settings have their initial values) or if they are calling it a second time, after changing some of the values.

So we have two options here -- either consistently not mention the initial values (the minimalist approach), or fill them in for each of the parameters (more helpful given that we don't currently have documentation for the frontend functions). Either way, I think we need to add a sentence to the docstring here:

``All parameters are optional. For each option that is not provided, the value currently set in the frontend will be preserved.``

If we do want to fill in all the initial values, we should then also add ``Initial frontend settings are noted below.``

Then for each of the parameters we should put something like `Initially set to X.` or `Initially disabled.`, and add `This default is configurable in the frontend preferences.` if that value is read from the prefs.

I will give more specific examples for the parameters in this function later today.

confluence (changes requested):

confluence:
> The `intensity_min` and `intensity_max` stay to the previous values (1 and 2 in this case) and I can't reset it to `Automatic`.

Are you sure you're testing with the updated frontend?

I can't see anything obviously incorrect in your implementation (I did request changes to the init parameters, but that should make no functional difference).

Please update the min/max logic in the function (the issue may be there), and also test the low-level action call both with a normal macro and with the new subclass.

carta/image.py
```diff
@@ -672,6 +672,225 @@ def hide_contours(self):
...
+            The rotation offset in degrees. By default is 0.
+        """
+        if thickness is not None:
+            self.call_action("vectorOverlayConfig.setThickness", thickness)
+        if intensity_min is None:
+            intensity_min = self.macro("vectorOverlayConfig", "intensityMin")
+        elif intensity_min is Auto.AUTO:
+            intensity_min = Undefined()
+        else:
+            intensity_min = intensity_min
```
confluence: This is unnecessary -- you're setting the variable to its own value.

carta/image.py
```diff
@@ -672,6 +672,225 @@ def hide_contours(self):
...
+        if intensity_max is None:
+            intensity_max = self.macro("vectorOverlayConfig", "intensityMax")
+        elif intensity_max is Auto.AUTO:
+            intensity_max = Undefined()
+        else:
+            intensity_max = intensity_max
```
confluence: As above.

confluence (changes requested):

carta/image.py
I-Chenn: Done in 142c3b7.

carta/image.py
I-Chenn: Done in 142c3b7.

carta/image.py
I-Chenn: Done in 142c3b7.

carta/image.py
I-Chenn: Done in 142c3b7.

carta/image.py
I-Chenn: Done in 142c3b7.

carta/image.py
I-Chenn: Done in 142c3b7.

carta/image.py
I-Chenn: Done in 142c3b7.

carta/image.py
I-Chenn: Done in 142c3b7.

carta/image.py
I-Chenn: Done in 142c3b7.

carta/image.py
kswang1029: @I-Chenn what's the idea behind of having this kind of statement (`a = a`)? Please elaborate.

confluence:
~The changes from your last commit are not showing up in the default PR diff view -- I can't figure out exactly why; this seems to be a GitHub quirk. Could you please try merging dev into your branch and pushing it?~ -- never mind; I realised that the one line that I kept seeing unchanged was the line that you forgot to change in the commit. The diff is actually fine.

carta/image.py
I-Chenn: With the intention of listing all the possible values to clarify the logic of the code for developers/users, though I know it's stupid to set a variable to its own value. Sorry for that.

carta/image.py
confluence: This one change wasn't done in this commit.

confluence: Just a note: you don't need to reply to each review comment with a link to the commit -- I can see the commit in the diff, and I can compare the before/after if I need to.

I-Chenn:
> ~The changes from your last commit are not showing up in the default PR diff view -- I can't figure out exactly why; this seems to be a GitHub quirk. Could you please try merging dev into your branch and pushing it?~ -- never mind; I realised that the one line that I kept seeing unchanged was the line that you forgot to change in the commit. The diff is actually fine.

Oh, I see. May I know which line is that?

carta/image.py
confluence: Updated suggestion for the additional paragraphs:
```
All parameters are optional. For each option that is not provided, the value currently set in the frontend will be preserved. Initial frontend settings are noted below.

We deduce some boolean options. For example, providing an explicit pixel averaging width with the **pixel_averaging** parameter will automatically enable pixel averaging unless **pixel_averaging_enabled** is also explicitly set to ``False``. To disable pixel averaging, explicitly set **pixel_averaging_enabled** to ``False``.
```

carta/image.py
```diff
@@ -672,6 +672,221 @@ def hide_contours(self):
...
+        pixel_averaging_enabled : {2}
+            Enable pixel averaging.
+        pixel_averaging : {3}
+            The pixel averaging width in pixel.
+        fractional_intensity : {4}
+            Enable fractional polarization intensity. By default the absolute polarization intensity is used.
+        threshold_enabled : {5}
+            Enable threshold.
+        threshold : {6}
+            The threshold in Jy/pixels.
```
confluence: Here's an example of how the initial values should be written up. If you look up these properties in the vector overlay config store, you should see that they are read from the preferences (averaging is enabled if the pixel value is positive). If you look in the preference store, the default pixel value there is 4. So putting this all together, these descriptions should say:
```
pixel_averaging_enabled : {2}
    Enable pixel averaging. This is initially enabled if the pixel averaging width is positive.
pixel_averaging : {3}
    The pixel averaging width in pixels. The initial value can be configured in the frontend preferences (the default is ``4``).
```

(It's fine to use the word default in that second line because it's clear from the context that we're talking about the default of the frontend preference.)

Similarly, fractional intensity is configurable in the preferences and is disabled by default, so that description should be rewritten as:
```
fractional_intensity : {4}
    Enable fractional polarization intensity. The initial value can be configured in the frontend preferences. By default this is disabled and the absolute polarization intensity is used.
```

The initial threshold options are not configurable; they're just hardcoded in the store constructor:
```
threshold_enabled : {5}
     Enable threshold. Initially the threshold is disabled.
threshold : {6}
     The threshold in Jy/pixels. The initial value is zero.
```

carta/image.py
```diff
@@ -672,6 +672,221 @@ def hide_contours(self):
...
+        debiasing : {7}
+            Enable debiasing.
+        q_error : {8}
+            The Stokes Q error in Jy/beam. Set both this and ``u_error`` to enable debiasing. Debiasing is disabled by default.
+        u_error : {9}
+            The Stokes U error in Jy/beam. Set both this and ``q_error`` to enable debiasing. Debiasing is disabled by default.
```
confluence: The Q and U errors aren't actually initialized in the frontend (something which should perhaps be fixed there). I suspect that if they're unset the protobuf message values default to zero -- ~I'm checking this in the backend.~ So this should say something like:
```
debiasing : {7}
    Enable debiasing. This is initially disabled.
q_error : {8}
    The Stokes Q error in Jy/beam. Set both this and ``u_error`` to enable debiasing. Initially set to zero.
u_error : {9}
    The Stokes U error in Jy/beam. Set both this and ``q_error`` to enable debiasing. Initially set to zero.
```

~The effective default error values can be filled in later.~ I checked, and confirmed that if the frontend sets these to "undefined" in the message they're read as zeros in the backend.

I am explicitly mentioning setting both of these to enable debiasing, because it's a special case (both values are required). I'm not mentioning it for all the other similar cases because I think they're adequately covered by the general paragraph I suggested above.

confluence: You should follow a similar process for the color and colormap functions, and once all these descriptions are correct the `plot_vector_overlay` description should be updated to match.

confluence (changes requested):

confluence:
> Oh, I see. May I know which line is that?

It's the intensity source description. I just posted a few more review comments, including examples of how the initial values should be described.

I-Chenn:
@confluence 
I haven't checked for the HTML color for GREEN3, but I need to be away from the computer for a few hours. I will complete it today.

I-Chenn:
@confluence, all the descriptions for `vector_overlay` functions are optimized, and I also added the descripted default value for `color` and `colormap` for `contour`. Please take a look and see if it's OK for you.

confluence (approved):
Clearing my own review.

confluence:
@kswang1029 this is now ready for review.

(I would like to merge this first, and finalise making the raster / vector overlay / contours interface consistent in #123.)

PR #93, opened 2023-03-27
Icchen/75 single multiple panel

### LINKED ISSUE #75: support customization of the viewer modes: single-panel view and multi-panel view
as part of the epic https://github.com/CARTAvis/carta-python/issues/42, we will need a function to control the layout of the image viewer. Currently CARTA supports single-panel view and multi-panel view. There is a button at the top-right corner of the widget for switching the two modes. 
Screen Shot 2023-03-07 at 19 08 59

If it is the multi-panel view mode, and there are more images loaded (eg 5) than the configured grid layout (default 2x2), we can also use the adjacent buttons to switch to other "pages".

In the settings dialog, we can configure the grid layouts by setting n_row and n_column.

In short, we need a function or functions to configure the grid layout of the image viewer.

I-Chenn:
This PR is to resolve #75.

Four functions: `set_viewer_mode`, `previous_page`, `next_page` and `set_viewer_grid` are added to `session.py`.
`set_viewer_mode`: Switch between single-mode and multiple-mode.
`previous_page`: Click "previous page" button
`previous_page`: Click "next page" button
`set_viewer_grid`: Set "Enable multi-panel" to `True` and define the number of columns and/or rows.

### DIFF
--- a/carta/session.py
+++ b/carta/session.py
@@ -427,12 +427,15 @@ class Session:
         Parameters
         ----------
         panel : {0}
-            To use single-panel mode (``single``) or multiple-panel mode (``multiple``). Default is ``single``.
+            To adopt single-panel mode (``single``) or multiple-panel mode (``multiple``). Default is ``single``.
         """
         if panel is "single":
             if_multiple = False
         elif panel is "multiple":
             if_multiple = True
+        else:
+            logger.warning('Cannot proceed with the input panel argument. Please set "single" or "multiple" for the argument.')
+            return
         self.call_action("widgetsStore.setImageMultiPanelEnabled", if_multiple)
 
     def previous_page(self):
@@ -443,8 +446,8 @@ class Session:
         """Go to next page in viewer."""
         self.call_action("widgetsStore.onNextPageClick")
 
-    @validate(Number(), Number())
-    def set_viewer_grid(self, columns=2, rows=2):
+    @validate(OneOf(*range(1, 101)), OneOf(*range(1, 101)), String())
+    def set_viewer_grid(self, columns=2, rows=2, mode="dynamic"):
         """
         Set number of columns and rows in viewer grid.
 
@@ -454,10 +457,17 @@ class Session:
             Number of columns. Default is 2.
         rows : {1}
             Number of rows. Default is 2.
+        mode : {2}
+            To adopt dynamic grid size mode (``dynamic``) or fixed grid size mode(``fixed``). ``dynamic`` is set as default.
         """
         self.call_action("widgetsStore.setImageMultiPanelEnabled", True)
         self.call_action("preferenceStore.setPreference", "imagePanelColumns", columns)
         self.call_action("preferenceStore.setPreference", "imagePanelRows", rows)
+        if mode is "dynamic" or mode is "fixed":
+            self.call_action("preferenceStore.setPreference", "imagePanelMode", mode)
+        else:
+            logger.warning('Cannot proceed with the input mode argument. Please set "dynamic" or "fixed" for the argument')
+            return
 
     # CANVAS AND OVERLAY
     @validate(Number(), Number())
--- a/carta/session.py
+++ b/carta/session.py
@@ -417,6 +417,11 @@ class Session:
         """Clear the raster scaling reference."""
         self.call_action("clearRasterScalingReference")
 
+    # VIEWER MODES
+    @validate(Boolean())
+    def set_viewer_mode(self, if_multiple=True):
+        self.call_action("widgetsStore.setImageMultiPanelEnabled", if_multiple)
+
     # CANVAS AND OVERLAY
     @validate(Number(), Number())
     def set_view_area(self, width, height):
--- a/carta/session.py
+++ b/carta/session.py
@@ -420,13 +420,39 @@ class Session:
     # VIEWER MODES
     @validate(Boolean())
     def set_viewer_mode(self, if_multiple=True):
+        """
+        Switch between single-panel mode and multiple-panel mode.
+
+        Parameters
+        ----------
+        if_multiple : {0}
+            Argument to decide if multiple-mode is adopted. Default is True.
+        """
         self.call_action("widgetsStore.setImageMultiPanelEnabled", if_multiple)
 
     def previous_page(self):
+        """Go to previous page on viewer"""
         self.call_action("widgetsStore.onPreviousPageClick")
 
     def next_page(self):
+        """GO to next page on viewer"""
         self.call_action("widgetsStore.onNextPageClick")
+
+    @validate(Number(), Number())
+    def set_viewer_grid(self, n_column=2, n_row=2):
+        """
+        Set numbers of columns and rows for viewer grids.
+
+        Parameters
+        ----------
+        n_column : {0}
+            Number of columns; Default is 2.
+        n_row : {1}
+            Number of rows; Default is 2.
+        """
+        self.call_action("widgetsStore.setImageMultiPanelEnabled", True)
+        self.call_action("preferenceStore.setPreference", "imagePanelColumns", n_column)
+        self.call_action("preferenceStore.setPreference", "imagePanelRows", n_row)
 
     # CANVAS AND OVERLAY
     @validate(Number(), Number())
--- a/carta/session.py
+++ b/carta/session.py
@@ -9,7 +9,7 @@ Alternatively, the user can create a new session which runs in a headless browse
 import base64
 
 from .image import Image
-from .constants import CoordinateSystem, LabelType, BeamType, PaletteColor, Overlay
+from .constants import CoordinateSystem, LabelType, BeamType, PaletteColor, Overlay, PanelMode, GridMode
 from .backend import Backend
 from .protocol import Protocol
 from .util import logger, Macro, split_action_path, CartaBadID, CartaBadSession, CartaBadUrl
@@ -419,21 +419,21 @@ class Session:
         self.call_action("clearRasterScalingReference")
 
     # VIEWER MODES
-    @validate(OneOf(["single", "multiple"]))
-    def set_viewer_mode(self, mode):
+    @validate(Constant(PanelMode))
+    def set_viewer_mode(self, panel_mode):
         """
         Switch between single-panel mode and multiple-panel mode.
 
         Parameters
         ----------
-        panel : {0}
-            To adopt single-panel mode (``single``) or multiple-panel mode (``multiple``).
+        panel_mode : {0}
+            To adopt single-panel mode (``0``) or multiple-panel mode (``1``).
         """
-        if mode is "single":
-            if_multiple = False
-        elif mode is "multiple":
-            if_multiple = True
-        self.call_action("widgetsStore.setImageMultiPanelEnabled", if_multiple)
+        if panel_mode == PanelMode.SINGLE.value:
+            multiple = False
+        elif panel_mode == PanelMode.MULTIPLE.value:
+            multiple = True
+        self.call_action("widgetsStore.setImageMultiPanelEnabled", multiple)
 
     def previous_page(self):
         """Go to previous page in viewer."""
@@ -443,8 +443,8 @@ class Session:
         """Go to next page in viewer."""
         self.call_action("widgetsStore.onNextPageClick")
 
-    @validate(OneOf(*range(1, 11)), OneOf(*range(1, 11)), OneOf(["dynamic", "fixed"]))
-    def set_viewer_grid(self, rows, columns, mode="fixed"):
+    @validate(OneOf(*range(1, 11)), OneOf(*range(1, 11)), Constant(GridMode))
+    def set_viewer_grid(self, rows, columns, grid_mode=GridMode.FIXED.value):
         """
         Set number of columns and rows in viewer grid.
 
@@ -454,13 +454,13 @@ class Session:
             Number of rows.
         columns : {1}
             Number of columns.
-        mode : {2}
+        grid_mode : {2}
             To adopt dynamic grid size mode (``dynamic``) or fixed grid size mode(``fixed``). Default is ``fixed``.
         """
         self.call_action("widgetsStore.setImageMultiPanelEnabled", True)
         self.call_action("preferenceStore.setPreference", "imagePanelRows", rows)
         self.call_action("preferenceStore.setPreference", "imagePanelColumns", columns)
-        self.call_action("preferenceStore.setPreference", "imagePanelMode", mode)
+        self.call_action("preferenceStore.setPreference", "imagePanelMode", grid_mode)
 
     # CANVAS AND OVERLAY
     @validate(Number(), Number())
--- a/carta/session.py
+++ b/carta/session.py
@@ -426,7 +426,7 @@ class Session:
         Parameters
         ----------
         if_multiple : {0}
-            Argument to decide if multiple-mode is adopted. Default is True.
+            Whether to use multiple-panel mode. By default is ``True``.
         """
         self.call_action("widgetsStore.setImageMultiPanelEnabled", if_multiple)
 
@@ -435,24 +435,24 @@ class Session:
         self.call_action("widgetsStore.onPreviousPageClick")
 
     def next_page(self):
-        """GO to next page on viewer"""
+        """Go to next page on viewer"""
         self.call_action("widgetsStore.onNextPageClick")
 
     @validate(Number(), Number())
-    def set_viewer_grid(self, n_column=2, n_row=2):
+    def set_viewer_grid(self, n_columns=2, n_rows=2):
         """
         Set numbers of columns and rows for viewer grids.
 
         Parameters
         ----------
         n_column : {0}
-            Number of columns; Default is 2.
+            Number of columns. By default is 2.
         n_row : {1}
-            Number of rows; Default is 2.
+            Number of rows. By default is 2.
         """
         self.call_action("widgetsStore.setImageMultiPanelEnabled", True)
-        self.call_action("preferenceStore.setPreference", "imagePanelColumns", n_column)
-        self.call_action("preferenceStore.setPreference", "imagePanelRows", n_row)
+        self.call_action("preferenceStore.setPreference", "imagePanelColumns", n_columns)
+        self.call_action("preferenceStore.setPreference", "imagePanelRows", n_rows)
 
     # CANVAS AND OVERLAY
     @validate(Number(), Number())
--- a/carta/session.py
+++ b/carta/session.py
@@ -419,16 +419,16 @@ class Session:
 
     # VIEWER MODES
     @validate(Boolean())
-    def set_viewer_mode(self, if_multiple=True):
+    def set_viewer_mode(self, multiple=True):
         """
         Switch between single-panel mode and multiple-panel mode.
 
         Parameters
         ----------
-        if_multiple : {0}
+        multiple : {0}
             Whether to use multiple-panel mode. By default is ``True``.
         """
-        self.call_action("widgetsStore.setImageMultiPanelEnabled", if_multiple)
+        self.call_action("widgetsStore.setImageMultiPanelEnabled", multiple)
 
     def previous_page(self):
         """Go to previous page on viewer"""
--- a/carta/session.py
+++ b/carta/session.py
@@ -426,33 +426,33 @@ class Session:
         Parameters
         ----------
         multiple : {0}
-            Whether to use multiple-panel mode. By default is ``True``.
+            Whether to use multiple-panel mode. Default is ``True``.
         """
         self.call_action("widgetsStore.setImageMultiPanelEnabled", multiple)
 
     def previous_page(self):
-        """Go to previous page on viewer"""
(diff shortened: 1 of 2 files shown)


carta/session.py
```diff
@@ -417,6 +417,43 @@ def clear_raster_scaling_reference(self):
         """Clear the raster scaling reference."""
         self.call_action("clearRasterScalingReference")
 
+    # VIEWER MODES
+    @validate(Boolean())
+    def set_viewer_mode(self, if_multiple=True):
```
confluence: This parameter should just be called `multiple`.

carta/session.py
```diff
@@ -417,6 +417,43 @@ def clear_raster_scaling_reference(self):
...
+        """
+        Switch between single-panel mode and multiple-panel mode.
+
+        Parameters
+        ----------
+        if_multiple : {0}
+            Argument to decide if multiple-mode is adopted. Default is True.
```
confluence: Remove "Argument to decide...". Perhaps something like "Whether to use multiple-panel mode." Also format `True` with double backticks.

carta/session.py
```diff
@@ -417,6 +417,43 @@ def clear_raster_scaling_reference(self):
...
+    def previous_page(self):
+        """Go to previous page on viewer"""
+        self.call_action("widgetsStore.onPreviousPageClick")
+
+    def next_page(self):
+        """GO to next page on viewer"""
+        self.call_action("widgetsStore.onNextPageClick")
+
+    @validate(Number(), Number())
+    def set_viewer_grid(self, n_column=2, n_row=2):
```
confluence: These should just be called `columns` and `rows`.

carta/session.py
```diff
@@ -417,6 +417,43 @@ def clear_raster_scaling_reference(self):
...
+        """
+        Set numbers of columns and rows for viewer grids.
+
+        Parameters
+        ----------
+        n_column : {0}
+            Number of columns; Default is 2.
+        n_row : {1}
+            Number of rows; Default is 2.
```
confluence: Either change the semicolon to a full stop, or lower-case "default".

confluence (changes requested):

carta/session.py
I-Chenn: Modificaition is applied in 8a8bf10

carta/session.py
I-Chenn: Modificaition is applied in 8a8bf10

carta/session.py
I-Chenn: Modificaition is applied in 8a8bf10

carta/session.py
I-Chenn: Modification is applied in 289bfb2.

carta/session.py
```diff
@@ -417,6 +417,43 @@ def clear_raster_scaling_reference(self):
...
+    def previous_page(self):
+        """Go to previous page on viewer"""
+        self.call_action("widgetsStore.onPreviousPageClick")
+
+    def next_page(self):
+        """Go to next page on viewer"""
+        self.call_action("widgetsStore.onNextPageClick")
+
+    @validate(Number(), Number())
+    def set_viewer_grid(self, n_columns=2, n_rows=2):
```
confluence: Here I meant that you should rename the variables to `rows` and `columns` -- no `n_`. Also, the names in the description have not been updated.

carta/session.py
```diff
@@ -417,6 +417,43 @@ def clear_raster_scaling_reference(self):
...
+    # VIEWER MODES
+    @validate(Boolean())
+    def set_viewer_mode(self, multiple=True):
+        """
+        Switch between single-panel mode and multiple-panel mode.
+
+        Parameters
+        ----------
+        multiple : {0}
+            Whether to use multiple-panel mode. By default is ``True``.
```
confluence: The default sentence should either be `Default: something.`, `The default is something.`, or `Defaults to something.` Sometimes we do use `By default [full sentence here...]`, but "by" should not be used here. (This also applies to the other description.)

carta/session.py
```diff
@@ -417,6 +417,43 @@ def clear_raster_scaling_reference(self):
...
+        """
+        self.call_action("widgetsStore.setImageMultiPanelEnabled", multiple)
+
+    def previous_page(self):
+        """Go to previous page on viewer"""
```
confluence: Should be "in viewer"

carta/session.py
```diff
@@ -417,6 +417,43 @@ def clear_raster_scaling_reference(self):
...
+        self.call_action("widgetsStore.onPreviousPageClick")
+
+    def next_page(self):
+        """Go to next page on viewer"""
+        self.call_action("widgetsStore.onNextPageClick")
+
+    @validate(Number(), Number())
+    def set_viewer_grid(self, n_columns=2, n_rows=2):
+        """
+        Set numbers of columns and rows for viewer grids.
```
confluence: I suggest `Set number of columns and rows in viewer grid.`

confluence (changes requested):

carta/session.py
I-Chenn: Done in d44b668.

carta/session.py
I-Chenn: Done in d44b668.

carta/session.py
I-Chenn: Done in d44b668.

carta/session.py
I-Chenn: Done in d44b668.

carta/session.py
```diff
@@ -417,6 +417,43 @@ def clear_raster_scaling_reference(self):
...
+            Whether to use multiple-panel mode. Default is ``True``.
+        """
+        self.call_action("widgetsStore.setImageMultiPanelEnabled", multiple)
+
+    def previous_page(self):
+        """Go to previous page in viewer"""
+        self.call_action("widgetsStore.onPreviousPageClick")
+
+    def next_page(self):
+        """Go to next page in viewer"""
```
confluence: One last thing: each of these docstrings should have a full stop at the end.

confluence (changes requested):

carta/session.py
```diff
@@ -417,6 +417,43 @@ def clear_raster_scaling_reference(self):
...
+        self.call_action("widgetsStore.onNextPageClick")
+
+    @validate(Number(), Number())
+    def set_viewer_grid(self, columns=2, rows=2):
+        """
+        Set number of columns and rows for viewer grid.
```
confluence: "For" should be "in".

confluence (changes requested):

carta/session.py
I-Chenn: Done in 85cccaf.

carta/session.py
I-Chenn: Done in 85cccaf

confluence (approved):

carta/session.py
```diff
@@ -417,6 +417,43 @@ def clear_raster_scaling_reference(self):
         """Clear the raster scaling reference."""
         self.call_action("clearRasterScalingReference")
 
+    # VIEWER MODES
+    @validate(Boolean())
+    def set_viewer_mode(self, multiple=True):
```
kswang1029: The frontend implementation uses a boolean to switch in between the two viewer modes. This does not mean we should expose this boolean to users. set_viewer_mode(True) / set_viewer_mode(False) is not intuitive and explicitly expressing the purpose of the function. I suggest we have a keyword argument as mode with options "single" and "multiple". Under the scene we can still use the boolean for control.

carta/session.py
```diff
@@ -417,6 +417,43 @@ def clear_raster_scaling_reference(self):
...
+
+    def previous_page(self):
+        """Go to previous page in viewer."""
+        self.call_action("widgetsStore.onPreviousPageClick")
+
+    def next_page(self):
+        """Go to next page in viewer."""
+        self.call_action("widgetsStore.onNextPageClick")
+
+    @validate(Number(), Number())
```
kswang1029: validate columns/rows a a number is not sufficient. Try to input 0 or -1 to see what happens. These two arguments should be positive integer.

carta/session.py
```diff
@@ -417,6 +417,43 @@ def clear_raster_scaling_reference(self):
...
+    def set_viewer_grid(self, columns=2, rows=2):
```
kswang1029: this function sets the grid layout. We would also need to specify how the grid layout is interpreted. In the settings dialog, we can set the multi-panel mode as dynamic or fixed. In this set_view_grid function, I suggest we add another keyword argument as mode and having options as "dynamic" or "fixed".

kswang1029 (changes requested):

carta/session.py
I-Chenn: Done in ab3d8af (14f6ca5, actually). The original `if_multiple` argument is changed to `panel`.  The new `panel` argument is allowed to set `"single"` and `"multiple"`. If it's not these two inputs, a logger. warning will appear to notify the user to set the correct input.

carta/session.py
I-Chenn: Done with a tricky method in ab3d8af. I set the validation check to `OneOf(*range(1, 101))` for `columns` and `rows`. I'm not sure if it's possible for user to set up to 100 for columns and rows, but I think it should work for most of the cases. Please let me know if it's a proper way of solving this issue.

carta/session.py
I-Chenn: In ab3d8af, this request is solved by adding  a `mode` argument in `set_viewer_grid`. The `mode` is allowed to set to `"dynamic"` and `"fixed"`; if it's not these two inputs, an warning will be printed out to imply the user to set correct input.

carta/session.py
```diff
@@ -418,6 +418,57 @@ def clear_raster_scaling_reference(self):
         """Clear the raster scaling reference."""
         self.call_action("clearRasterScalingReference")
 
+    # VIEWER MODES
+    @validate(String())
+    def set_viewer_mode(self, panel="single"):
```
kswang1029: shouldn't we use "mode" instead of "panel"? Also, this "panel" has default as "single". This means we can call the function like `session.set_viewer_mode()`. This is not intuitive as with a set function, we should expect a user input for what to be set.

carta/session.py
```diff
@@ -418,6 +418,57 @@ def clear_raster_scaling_reference(self):
...
+        """
+        Switch between single-panel mode and multiple-panel mode.
+
+        Parameters
+        ----------
+        panel : {0}
+            To adopt single-panel mode (``single``) or multiple-panel mode (``multiple``). Default is ``single``.
+        """
+        if panel is "single":
+            if_multiple = False
```
kswang1029: I will let @confluence to comment on the naming convention about "if_multiple".

carta/session.py
```diff
@@ -418,6 +418,57 @@ def clear_raster_scaling_reference(self):
...
+    # VIEWER MODES
+    @validate(String())
+    def set_viewer_mode(self, panel="single"):
+        """
+        Switch between single-panel mode and multiple-panel mode.
+
+        Parameters
+        ----------
+        panel : {0}
+            To adopt single-panel mode (``single``) or multiple-panel mode (``multiple``). Default is ``single``.
```
kswang1029: The single-panel / multi-panel mode is stored as an implicit preference in CARTA. So on the GUI, the state of the panel mode is determined dynamically when a new session is created. I would remove "Default is single" here as in reality the mode is an implicit preference. Also as mentioned in the above comment, we should let users determine the mode when `set_viewer_mode` is called.

carta/session.py
```diff
@@ -418,6 +418,57 @@ def clear_raster_scaling_reference(self):
...
+    def previous_page(self):
+        """Go to previous page in viewer."""
+        self.call_action("widgetsStore.onPreviousPageClick")
+
+    def next_page(self):
+        """Go to next page in viewer."""
+        self.call_action("widgetsStore.onNextPageClick")
+
+    @validate(OneOf(*range(1, 101)), OneOf(*range(1, 101)), String())
+    def set_viewer_grid(self, columns=2, rows=2, mode="dynamic"):
```
kswang1029: better swap the order of columns and rows to fit the usual convention. ie rows first, then columns

carta/session.py
```diff
@@ -418,6 +418,57 @@ def clear_raster_scaling_reference(self):
...
+
+    def previous_page(self):
+        """Go to previous page in viewer."""
+        self.call_action("widgetsStore.onPreviousPageClick")
+
+    def next_page(self):
+        """Go to next page in viewer."""
+        self.call_action("widgetsStore.onNextPageClick")
+
+    @validate(OneOf(*range(1, 101)), OneOf(*range(1, 101)), String())
```
kswang1029: currently the frontend does not apply an upper limit (it should actually). I suggest we apply 10 here as an upper limit for now. Or to be consistent with the current behavior of the frontend, there is no upper limit in the validator. 

Can OneOf be used to validate "dynamic" and "fixed"? Then we don't need `if` checks.

carta/session.py
kswang1029: this is a set function so it is better not to have default values for the input arguments. Also, columns, rows, and mode are implicit preferences from the frontend point of view.

kswang1029 (changes requested):

carta/session.py
kswang1029: just a note for the future, there will be a new mode as "channel" for channel map view.

carta/session.py
kswang1029: I _think_ when a user call this function to define a grid layout, the mode needs to be "fixed" by default. This would be more consistent with matplotlib.pyplot.subplot(), for example.

carta/session.py
I-Chenn: @kswang1029, In e62f8c4:
1. `mode` is used instead of `panel` for the argument.
2. Remove the default setting of `mode`.
3. Adopt `OneOf(["single", "multiple"])` instead of `String()` for the validation of `mode`.

carta/session.py
I-Chenn: The description about the default values are removed in e62f8c4.

carta/session.py
I-Chenn: In e62f8c4:
1. The order between `rows` and `columns` is swapped.
2. Remove the default settings and the descriptions for all arguments _except_ for `mode` (set to `fixed`).

Nevertheless, just a quick question: as for the default value, shouldn't we do what we did for the "set contour" or "set vector overlay" issue, that we can execute the function without any parameter set (like: `session.set_viewer_grid()`) then get the result of the default setting of the frontend (2x2 grids, dynamic mode)?

carta/session.py
I-Chenn: Yes, `OneOf` can validate between strings. In e62f8c4 it is used for validation and I also used it in the `set_viewer_mode` function. Thanks a lot.

carta/session.py
```diff
@@ -418,6 +418,50 @@ def clear_raster_scaling_reference(self):
         """Clear the raster scaling reference."""
         self.call_action("clearRasterScalingReference")
 
+    # VIEWER MODES
+    @validate(OneOf(["single", "multiple"]))
```
confluence: I prefer not to use "magic" string literals like this. Instead, I suggest that you create a constant called `PanelMode` with attributes `SINGLE` and `MULTIPLE`. Since these string values aren't actually used as strings, it should have the default integer values (like e.g. `SmoothingMode`).

carta/session.py
confluence: It should be called `multiple` (no `if_`).

carta/session.py
```diff
@@ -418,6 +418,50 @@ def clear_raster_scaling_reference(self):
...
+    def previous_page(self):
+        """Go to previous page in viewer."""
+        self.call_action("widgetsStore.onPreviousPageClick")
+
+    def next_page(self):
+        """Go to next page in viewer."""
+        self.call_action("widgetsStore.onNextPageClick")
+
+    @validate(OneOf(*range(1, 11)), OneOf(*range(1, 11)), OneOf(["dynamic", "fixed"]))
+    def set_viewer_grid(self, rows, columns, mode="fixed"):
```
confluence: Similarly, this mode should be set with a constant (perhaps called `GridMode`). But here the string values are actually sent to the frontend, so the enum should be a string enum with those values (like e.g. `LabelType`).

confluence (changes requested):

carta/session.py
I-Chenn: In 1684398, `PanelMode` is added to `constants.py`:
截圖 2023-04-12 下午8 42 40
and is then used in `set_viewer_mode`:
截圖 2023-04-12 下午8 45 16

carta/session.py
I-Chenn: In 1684398, `GridMode` is added to `constants.py`:
截圖 2023-04-12 下午8 49 37
And is then used in `set_viewer_grid`:
截圖 2023-04-12 下午8 52 19

carta/session.py
```diff
@@ -418,6 +418,50 @@ def clear_raster_scaling_reference(self):
...
+        Switch between single-panel mode and multiple-panel mode.
+
+        Parameters
+        ----------
+        panel_mode : {0}
+            To adopt single-panel mode (``0``) or multiple-panel mode (``1``).
+        """
+        if panel_mode == PanelMode.SINGLE.value:
+            multiple = False
+        elif panel_mode == PanelMode.MULTIPLE.value:
```
confluence: You don't need to use an explicit `.value` anywhere; you can just use e.g. `PanelMode.SINGLE`. This applies to `GridMode` too.

carta/session.py
```diff
@@ -418,6 +418,50 @@ def clear_raster_scaling_reference(self):
...
+    # VIEWER MODES
+    @validate(Constant(PanelMode))
+    def set_viewer_mode(self, panel_mode):
+        """
+        Switch between single-panel mode and multiple-panel mode.
+
+        Parameters
+        ----------
+        panel_mode : {0}
+            To adopt single-panel mode (``0``) or multiple-panel mode (``1``).
```
confluence: Don't use numbers here; use the enum names (`PanelMode.SINGLE` and `PanelMode.MULTIPLE`). Refer to them with their full object paths; then they will be formatted and linked correctly in the docs. E.g.
```
:obj:`carta.constants.PanelMode.SINGLE`
```

carta/session.py
```diff
@@ -418,6 +418,50 @@ def clear_raster_scaling_reference(self):
...
+        Set number of columns and rows in viewer grid.
+
+        Parameters
+        ----------
+        rows : {0}
+            Number of rows.
+        columns : {1}
+            Number of columns.
+        grid_mode : {2}
+            To adopt dynamic grid size mode (``dynamic``) or fixed grid size mode(``fixed``). Default is ``fixed``.
```
confluence: Also use the enum symbols here, not the string values.

confluence (changes requested):

carta/session.py
confluence: (Omit the double backticks when you use this format.)

carta/session.py
```diff
@@ -455,7 +455,7 @@ def set_viewer_grid(self, rows, columns, grid_mode=GridMode.FIXED.value):
         columns : {1}
             Number of columns.
         grid_mode : {2}
-            To adopt dynamic grid size mode (``dynamic``) or fixed grid size mode(``fixed``). Default is ``fixed``.
+            To adopt dynamic grid size mode (:obj:`grid_mode=GridMode.DYNAMIC`) or fixed grid size mode(:obj:`grid_mode=GridMode.FIXED`). Default is :obj:`grid_mode=GridMode.FIXED`.
```
confluence: I think you pasted the wrong thing here -- should be the path.

carta/session.py
```diff
@@ -427,11 +427,11 @@ def set_viewer_mode(self, panel_mode):
         Parameters
         ----------
         panel_mode : {0}
-            To adopt single-panel mode (``0``) or multiple-panel mode (``1``).
+            To adopt single-panel mode (:obj:`carta.constants.PanelMode.SINGLE`) or multiple-panel mode (:obj:`carta.constants.PanelMode.MULTIPLE`).
```
confluence: Actually, because the enum will be linked from the type description, it isn't necessary to list the options here either. Also, the description should be a noun phrase. So I would suggest just `The panel mode to adopt.`

carta/session.py
confluence: I also think listing the options here isn't necessary, and this should be rephrased -- but it's fine to keep the default. So:
```
The grid mode to adopt. The default is :obj:`carta.constants.GridMode.FIXED`.
```

confluence (changes requested):

carta/session.py
I-Chenn: Done in dec7725.

carta/session.py
I-Chenn: Done in dec7725.

confluence (approved):
This looks good! One last thing -- I forgot to mention that there's a missing newline at the end of `constants.py`.

There are scripts in `scripts` for checking and fixing this and other formatting issues -- you should run them before commits (and I should set up a hook for it in this repo!). I will make the change in this branch before merging.

PR #96, opened 2023-03-28
Icchen/77 clip level for colormap mapping

### LINKED ISSUE #77: support a clip level for colormap mapping
Image.set_scaling() takes a min arg and a max arg as the range for colormap mapping. It would be handy if we support a clip arg in % so min and max are _computed_. If clip, min and max are all given by users, we ignore min and max.

I-Chenn:
This PR is to resolve #77.

An argument `rank` is added to the `set_scaling` function of `image.py`. `rank` (in %) is a used to call `renderConfig.setPercentileRank` at the frontend and calculate/set the `Clip min` and `Clip max` accordingly.

`clip` cannot be used with `min` and `max`, and vise versa. Since we're still having "`None` can not be serialized" problem, the "non-empty string" method is adopted initiatively to replace the `None` value.

### DIFF
--- a/carta/image.py
+++ b/carta/image.py
@@ -484,7 +484,7 @@ class Image:
             self.call_action("renderConfig.resetContrast")
 
     # TODO check whether this works as expected
-    @validate(Constant(Scaling), NoneOr(Number()), NoneOr(Number()), NoneOr(Number()), NoneOr(Number()), NoneOr(Number()))
+    @validate(Constant(Scaling), NoneOr(Number()), NoneOr(Number()), NoneOr(Number(0, 100)), NoneOr(Number()), NoneOr(Number()))
     def set_scaling(self, scaling, alpha=None, gamma=None, rank=None, min=None, max=None):
         """Set the colormap scaling.
 
@@ -702,7 +702,13 @@ class Image:
         rank : {0}
             The percentile rank.
         """
+        percentile_ranks = [90, 95, 99, 99.5, 99.9, 99.95, 99.99, 100]
         self.call_action("renderConfig.setPercentileRank", rank)
+        min_val = self.get_value("renderConfig.scaleMinVal")
+        max_val = self.get_value("renderConfig.scaleMaxVal")
+        if rank not in percentile_ranks:
+            self.call_action("renderConfig.setCustomScale", min_val, max_val)
+
 
     # CLOSE
 
--- a/carta/image.py
+++ b/carta/image.py
@@ -466,7 +466,7 @@ class Image:
 
     # TODO check whether this works as expected
     @validate(Constant(Scaling), NoneOr(Number()), NoneOr(Number()), NoneOr(Number()), NoneOr(Number()), NoneOr(Number()))
-    def set_scaling(self, scaling, alpha=None, gamma=None, clip=None, min=None, max=None):
+    def set_scaling(self, scaling, alpha=None, gamma=None, rank=None, min=None, max=None):
         """Set the colormap scaling.
 
         Parameters
@@ -477,24 +477,21 @@ class Image:
             The alpha value (only applicable to ``LOG`` and ``POWER`` scaling types).
         gamma : {2}
             The gamma value (only applicable to the ``GAMMA`` scaling type).
-        clip : {3}
-            The scale clip. In percentage. Can not be used witn *min* and *max*.
+        rank : {3}
+            The percentile rank. If this is set, *min* and *max* are ignored, and will be calculated automatically.
         min : {4}
-            The minimum of the scale. Only used if both *min* and *max* are set. Can not be used with *clip*.
+            Custom clip minimum. Only used if both *min* and *max* are set. Ignored if *rank* is set.
         max : {5}
-            The maximum of the scale. Only used if both *min* and *max* are set. Can not be used with *clip*.
+            Custom clip maximum. Only used if both *min* and *max* are set. Ignored if *rank* is set.
         """
         self.call_action("renderConfig.setScaling", scaling)
         if scaling in (Scaling.LOG, Scaling.POWER) and alpha is not None:
             self.call_action("renderConfig.setAlpha", alpha)
         elif scaling == Scaling.GAMMA and gamma is not None:
             self.call_action("renderConfig.setGamma", gamma)
-        if clip is not None:
-            min = "None"
-            max = "None"
-            self.call_action("renderConfig.setPercentileRank", clip)
+        if rank is not None:
+            self.set_percentile_rank(rank)
         elif min is not None and max is not None:
-            clip = "None"
             self.call_action("renderConfig.setCustomScale", min, max)
 
     @validate(Boolean())
--- a/carta/image.py
+++ b/carta/image.py
@@ -465,8 +465,8 @@ class Image:
             self.call_action("renderConfig.resetContrast")
 
     # TODO check whether this works as expected
-    @validate(Constant(Scaling), NoneOr(Number()), NoneOr(Number()), NoneOr(Number()), NoneOr(Number()))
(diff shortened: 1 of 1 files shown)


carta/image.py
```diff
@@ -477,17 +477,24 @@ def set_scaling(self, scaling, alpha=None, gamma=None, min=None, max=None):
...
-        min : {3}
-            The minimum of the scale. Only used if both *min* and *max* are set.
-        max : {4}
-            The maximum of the scale. Only used if both *min* and *max* are set.
+        clip : {3}
+            The scale clip. In percentage. Can not be used witn *min* and *max*.
+        min : {4}
+            The minimum of the scale. Only used if both *min* and *max* are set. Can not be used with *clip*.
+        max : {5}
+            The maximum of the scale. Only used if both *min* and *max* are set. Can not be used with *clip*.
```
confluence: These descriptions should be changed to:

 `The clip percentage. If this is set, *min* and *max* are ignored, and will be calculated automatically.`

`Custom clip minimum. Only used if both *min* and *max* are set. Ignored if *clip* is set.`

`Custom clip maximum. Only used if both *min* and *max* are set. Ignored if *clip* is set.`

carta/image.py
```diff
@@ -477,17 +477,24 @@ def set_scaling(self, scaling, alpha=None, gamma=None, min=None, max=None):
...
         elif scaling == Scaling.GAMMA and gamma is not None:
             self.call_action("renderConfig.setGamma", gamma)
-        if min is not None and max is not None:
+        if clip is not None:
+            min = "None"
+            max = "None"
+            self.call_action("renderConfig.setPercentileRank", clip)
+        elif min is not None and max is not None:
+            clip = "None"
             self.call_action("renderConfig.setCustomScale", min, max)
```
confluence: 1. You are not actually using these `"None"` strings after creating them, because neither block in the conditional is calling the other action.
2. It is unnecessary for you to unset these values in the frontend, because the frontend functions already do that. `setPercentileRank` automatically sets the clip min and max, and vice versa. You should remove these three lines entirely.

confluence: There is already a high-level function called `set_percentile_rank`, which changes only the rank -- it should probably be used here instead of the low-level action call.

confluence (changes requested):

carta/image.py
I-Chenn: Thanks for the reminding of the `set_percentile_rank`, I didn't know the existence of that function. I used it in the `set_scaling` instead of action call in 7b81303. To avoid confusion, I thus change the argument `clip` to `rank`.

I-Chenn: I remove these three lines in 7b81303.

carta/image.py
I-Chenn: Done in 7b81303.

To avoid confusion, I also replace the argument from `clip` to `rank` to unify the name of the argument with `set_percentile_rank`. The description of `rank` is thus modified the replacement:

```
rank : {3}
    The percentile rank. If this is set, *min* and *max* are ignored, and will be calculated automatically.
min : {4}
    Custom clip minimum. Only used if both *min* and *max* are set. Ignored if *rank* is set.
max : {5}
    Custom clip maximum. Only used if both *min* and *max* are set. Ignored if *rank* is set.
```

carta/image.py
```diff
@@ -477,17 +477,21 @@ def set_scaling(self, scaling, alpha=None, gamma=None, min=None, max=None):
             The alpha value (only applicable to ``LOG`` and ``POWER`` scaling types).
         gamma : {2}
             The gamma value (only applicable to the ``GAMMA`` scaling type).
-        min : {3}
-            The minimum of the scale. Only used if both *min* and *max* are set.
-        max : {4}
-            The maximum of the scale. Only used if both *min* and *max* are set.
+        rank : {3}
+            The percentile rank. If this is set, *min* and *max* are ignored, and will be calculated automatically.
```
confluence: Renaming this for consistency is a good idea, but I think the complete absence of the word "clip" makes the naming of this set of parameters a little confusing. I would add it in this line only: `The clip percentile rank...`.

I think that for clarity and consistency we should also edit the description of `set_percentile_rank` so that the first line is `Set the clip percentile.`. The function should perhaps be called `set_clip_percentile` (the parameter would still be called `rank`). This would change the API, and would affect any existing code. @kswang1029 , what do you think?

kswang1029: ok

I-Chenn: In 48db542, the `set_percentile_rank` is renamed with `set_clip_percentile`. The descriptions are modified according to @confluence's suggestions.

confluence (approved):

carta/image.py
```diff
@@ -465,8 +465,8 @@ def set_colormap(self, colormap, invert=False, bias=None, contrast=None):
             self.call_action("renderConfig.resetContrast")
 
     # TODO check whether this works as expected
-    @validate(Constant(Scaling), NoneOr(Number()), NoneOr(Number()), NoneOr(Number()), NoneOr(Number()))
-    def set_scaling(self, scaling, alpha=None, gamma=None, min=None, max=None):
+    @validate(Constant(Scaling), NoneOr(Number()), NoneOr(Number()), NoneOr(Number()), NoneOr(Number()), NoneOr(Number()))
```
kswang1029: we should validate rank as a number between 0 and 100 (same validation as `set_clip_percentile`).

carta/image.py
```diff
@@ -630,8 +634,8 @@ def use_channel_histogram(self, contours=False):
         self.call_action(f"renderConfig.setUseCubeHistogram{'Contours' if contours else ''}", False)
 
     @validate(Number(0, 100))
-    def set_percentile_rank(self, rank):
-        """Set the percentile rank.
+    def set_clip_percentile(self, rank):
```
kswang1029, line 697: There is an existing issue about setting the rank not in the preset ranks (90, 95, 99, ...). If we set the rank as 90, the 90% button will be selected (down). But if we set a rank such as 80, no button will be selected. I suggest we select the "Custom" button in such cases. This means we trigger the "Custom" button _after_ we set the clip percentile.

kswang1029 (changes requested):

carta/image.py
I-Chenn: Done in 8f98ffd.

carta/image.py
I-Chenn: @kswang1029 @confluence 
Done in 8f98ffd. Check the result below:
https://drive.google.com/file/d/1N9b4mAfRp5Zqa6aJmX5LDoiNmxK2eKzv/view?usp=sharing

I-Chenn: @kswang1029, @confluence 
A minor modification is made in e4a6262, according to the suggestions of @YuHsuan-Hwang. `setPercentileRank` is called again to just "press the `Custom` button" instead of resetting the colormap with `setCustomScale`.

kswang1029 (approved):

carta/image.py
```diff
@@ -690,15 +694,18 @@ def use_channel_histogram(self, contours=False):
...
 
         Parameters
         ----------
         rank : {0}
             The percentile rank.
         """
+        percentile_ranks = [90, 95, 99, 99.5, 99.9, 99.95, 99.99, 100]
         self.call_action("renderConfig.setPercentileRank", rank)
+        if rank not in percentile_ranks:
+            self.call_action("renderConfig.setPercentileRank", -1)
```
confluence: This is a minor nitpick, but I would suggest renaming this variable to `preset_ranks`, and also adding an inline comment to the last line (something like `# select 'custom' rank button`) -- to make the intention here clear, because it's not obvious from the function name.

confluence (changes requested):

carta/image.py
I-Chenn: Done in 49f0887.

confluence (approved):

Issue #91, opened 2023-03-22
support loading image via LEL


kswang1029:
CARTA v3-stable supports loading an image via LEL with which users can perform arithmetic on multiple images and load the resulting image for visualization and perform further image analytics. 

With the scripting interface, we would need to support LEL too. Possibly we can have a `Session.open_LEL_image()` function with args as `expression=Str()`. Hopefully the carta_backend can handle errors in the expression. Example,

```
session = Session.start_and_create(...)
img = session.load_LEL_image("'image_a.fits' + 'image_b.fits'")
```

Issue #90, opened 2023-03-22
support loading a Stokes hypercube


kswang1029:
With the CARTA GUI, we can select 2-4 individual Stokes images to form a Stokes hypercube.

![Screen Shot 2023-03-22 at 12 27 29](https://user-images.githubusercontent.com/20819712/226802192-ef081465-6660-4b3d-b493-0c9ed946db60.png)

 In the scripting interface, we would need to support this feature too. As in the GUI there is an extra step to do file to Stokes components mapping, I suggest we make a new function like `Session.load_Stokes_hypercube()` with args as
`paths` (a list of images) and `Stokes_mapping` (a list of Stokes components). For example,

```
session = Session.start_and_create(...)
img = session.load_Stokes_hypercube(paths=["M51_band3_I.fits", "M51_band3_Q.fits", "M51_band3_U.fits"], Stokes_mapping=["I", "Q", "U"])
```

Issue #89, opened 2023-03-22
support loading a complex-valued image


kswang1029:
with `Session.open_image()` we can load a complex-valued image with the default "Amplitude".
There are other options as Amplitude, Phase, Real, and Imaginary. We would need to enhance open_image() to support the others.
![Screen Shot 2023-03-22 at 12 17 16](https://user-images.githubusercontent.com/20819712/226800966-609ac6d6-f658-4ab0-ab09-532b8438fe1f.png)

PR #98, opened 2023-03-29
Icchen/76 load multiple images

### LINKED ISSUE #76: generalizing Session.load_image() and Session.append_image() to allow loading multiple images in one call
Currently Session.load_image() and Session.append_image() accept a path arg to load one image. Although users can create a list of path and use a loop to call Session.append_image() to load multiple images, it would be handy if we can generalize the path as a str or a list of str so users can load images in one function call.

I-Chenn:
This PR is to resolve #76.

A function called `open_images` is added to `session.py` to open multiple images with one single funcition. User simply need to put the desired open images paths into the arguments of the function (in string), the function will then turn these paths into a list and open the images accordingly.

### DIFF
--- a/carta/session.py
+++ b/carta/session.py
@@ -383,6 +383,19 @@ class Session:
         """
         return Image.new(self, path, hdu, True)
 
+    def open_images(self, *paths):
+        """Open multiple images.
+
+        Parameters
+        ----------
+        paths : {0}
+            The paths to the image files, either relative to the session's current directory or an absolute path relative to the CARTA backend's root directory. Separate them with comma.
+        """
+        path_list = list(paths)
+        self.open_image(path_list[0])
+        for path in path_list[1:]:
+            self.append_image(path)
+
     def image_list(self):
         """Return the list of currently open images.
 


carta/session.py
```diff
@@ -383,6 +383,19 @@ def append_image(self, path, hdu=""):
         """
         return Image.new(self, path, hdu, True)
 
+    def open_images(self, *paths):
+        """Open multiple images.
+
+        Parameters
+        ----------
+        paths : {0}
```
confluence, line 400: You can't use `{0}` here -- that's only for substituting types automatically inserted by the validation decorator. And you shouldn't use that decorator here because you're using `*args` -- a variable-length iterable of parameters. This isn't supported by the validation decorators. You should manually write `iterable of strings` here.

The lack of validation doesn't matter because you're calling `open_image` and `append_image`, which use their own validation.

confluence: Also write `*paths` rather than `paths` in the description, for consistency with the way variable parameters are documented elsewhere in the code.

carta/session.py
```diff
@@ -383,6 +383,19 @@ def append_image(self, path, hdu=""):
...
+            The paths to the image files, either relative to the session's current directory or an absolute path relative to the CARTA backend's root directory. Separate them with comma.
```
confluence, line 401: `an absolute path` -> `absolute paths`

This is *not* a comma-separated string; it's an iterable of strings. Separate parameters passed to this function will be stored in a single tuple parameter called `paths`. (Look up `python args kwargs` to see how these "magic" variables work -- the names are conventions; it's fine to use the name `paths` here.)

carta/session.py
```diff
@@ -383,6 +383,19 @@ def append_image(self, path, hdu=""):
...
+        """
+        path_list = list(paths)
```
confluence, line 403: If this were a comma-separated string, this would not do the right thing -- it would split the string into a list of individual characters. But this is in fact a tuple of strings, so this line is unnecessary -- you can do everything with a tuple below that you can do with a list, so the cast is not required. You'd only have to do that if you needed to modify the list in-place.

Use `paths` directly below.

carta/session.py
```diff
@@ -383,6 +383,19 @@ def append_image(self, path, hdu=""):
         """
         return Image.new(self, path, hdu, True)
 
+    def open_images(self, *paths):
+        """Open multiple images.
```
confluence, line 396: This description should mention that any images already open in the session will be closed.

carta/session.py
```diff
@@ -383,6 +383,19 @@ def append_image(self, path, hdu=""):
...
+        Parameters
+        ----------
+        paths : {0}
+            The paths to the image files, either relative to the session's current directory or an absolute path relative to the CARTA backend's root directory. Separate them with comma.
+        """
+        path_list = list(paths)
+        self.open_image(path_list[0])
+        for path in path_list[1:]:
+            self.append_image(path)
+
```
confluence, line 407: This method should return a list of image objects which reference the opened images, for consistency with `open_image` and `append_image`. So you should create an empty `images` list, and append the values returned by these methods to that list. You can then return the list at the end.

carta/session.py
```diff
@@ -383,6 +383,19 @@ def append_image(self, path, hdu=""):
         """
         return Image.new(self, path, hdu, True)
 
+    def open_images(self, *paths):
```
confluence, line 395: This function will always open the default HDU for each image -- it's currently not possible to specify different HDUs. @kswang1029 will that be a problem? There are various ways that we could add this option, although they're all a bit messy. Personally I think that it's not difficult for the user to write their own loop. How often is it likely to come up?

This function will also always close any existing images. Should we also create an `append_images` function which does exactly the same thing except that it *doesn't* close existing images? This would just mean replacing the first `open_image` call with `append_image`. Instead of a separate function we could also add an option to this one.

confluence (changes requested):

kswang1029 (changes requested):
need to take multiple HDUs into account if it is a FITS image. there are other kinds of images to be supported too (eg complex image, etc)

carta/session.py
confluence: FWIW, the simplest and least messy way to allow optional specification of HDUs is to let the user pass in either strings or (string, string) tuples. Then in the function we'd check for this and call the underlying functions appropriately. At the cost of slightly more magic inside the function we could keep the simple API for the most common case (i.e. just plain strings for the default HDUs).

kswang1029:
I suggest to postpone the implementation of this feature for now. #89 #90 #91 should go first. Then we can have a better view of the structures to handle multiple images.

confluence:
This feature has been incorporated into #106, so I am closing this PR.

PR #100, opened 2023-04-10
Icchen/89 load complex image

### LINKED ISSUE #89: support loading a complex-valued image

I-Chenn:
This PR is to resolve #89.

@confluence, I bump into some problems when dealing with this issue:
I found that I can open a composite image with the script like:
```
img = session.call_action("openFile", "/media/icchen/disk/fits/", 'PHASE("complex.image")', "", "true", return_path="frameInfo.fileId")
```
Note that the `'PHASE("complex. Image")'` is at the position of `file_name`. Thus, in 2505e12, I added an argument `composite` to `new` of `image.py` and `open_image` and `append_image` of `session.py` to create new `file_name` if `composite` is not `None` (I haven't write the description for the argument, just make a test first).Then I tried to write a test script like:
```
img = session.open_image("/media/icchen/disk/fits/complex.image", composite="PHASE")
```
but got an message like:
```
-build/venv_carta-python/bin/python /home/icchen/Python_wrappers/test.py
  File "/home/icchen/Python_wrappers/test.py", line 16, in <module>
    img = session.open_image("/media/icchen/disk/fits/complex.image", composite="PHASE")
  File "/home/icchen/carta-build/venv_carta-python/lib/python3.10/site-packages/carta/validation.py", line 634, in newfunc
    return func(self, *args, **kwargs)
  File "/home/icchen/carta-build/venv_carta-python/lib/python3.10/site-packages/carta/session.py", line 372, in open_image
    return Image.new(self, path, hdu, False, composite)
  File "/home/icchen/carta-build/venv_carta-python/lib/python3.10/site-packages/carta/image.py", line 70, in new
    image_id = session.call_action("appendFile" if append else "openFile", directory, file_name, hdu, return_path="frameInfo.fileId")
  File "/home/icchen/carta-build/venv_carta-python/lib/python3.10/site-packages/carta/session.py", line 254, in call_action
    return self._protocol.request_scripting_action(self.session_id, path, *args, **kwargs)
  File "/home/icchen/carta-build/venv_carta-python/lib/python3.10/site-packages/carta/protocol.py", line 372, in request_scripting_action
    raise CartaActionFailed(f"{carta_action_description} failed: {response_data.get('message', 'No error message received.')}")
carta.util.CartaActionFailed: CARTA scripting action .openFile called with parameters ('/media/icchen/disk/fits', 'PHASE("complex. Image")', '') failed: File PHASE("complex.image") does not exist.
```
Note that in the last line it says
```
CARTA scripting action .openFile called with parameters ('/media/icchen/disk/fits', 'PHASE("complex.image")', '') failed: File PHASE("complex.image") does not exist.
```
but `PHASE("complex.image")` is exactly what I input with the script that I successfully open the composite image:
```
img = session.call_action("openFile", "/media/icchen/disk/fits/", 'PHASE("complex. Image")', "", "true", return_path="frameInfo.fileId")
```
Do you have idea why this `composite` string I create is not able to be recognized as a file when using `open_image`?

p.s. I have recorded a video for your reference:
[Screencast from 04-10-2023 12:19:51 PM.webm](https://user-images.githubusercontent.com/106953609/230826543-beae70b6-f8de-4751-ba7e-566b93a2c03e.webm)

### DIFF
--- a/carta/image.py
+++ b/carta/image.py
@@ -56,9 +56,9 @@ class Image:
         hdu : string
             The HDU to open.
         append : boolean
-            Whether the image should be appended. By default it is not, and all other open images are closed.
+            Whether the image should be appended.
         complex : a member of :obj:`carta.constants.ArithmeticExpression` or ``None``
-            Arithmetic expression to use if opening a complex-valued image. By default the image is assumed not to be complex.
+            Arithmetic expression to use if opening a complex-valued image. Set to ``None`` if the image is not complex, or to show the default amplitude if the image is complex.
 
         Returns
         -------
--- a/carta/image.py
+++ b/carta/image.py
@@ -42,7 +42,7 @@ class Image:
         self._frame = Macro("", self._base_path)
 
     @classmethod
-    def new(cls, session, path, hdu, append, complex=False, Expression=None):
+    def new(cls, session, path, hdu, append, complex, expression):
         """Open or append a new image in the session and return an image object associated with it.
 
         This method should not be used directly. It is wrapped by :obj:`carta.session.Session.open_image` and :obj:`carta.session.Session.append_image`.
@@ -58,9 +58,9 @@ class Image:
         append : boolean
             Whether the image should be appended.
         complex : boolean
-            Whether the image is complex. Set to ``False`` by default.
-        expression : a member of :obj:`carta.constants.ArithmeticExpression` or ``None``
-            Arithmetic expression to use if opening a complex-valued image. Set to ``None`` if the image is not complex, or to show the default amplitude if the image is complex.
+            Whether the image is complex.
+        expression : a member of :obj:`carta.constants.ArithmeticExpression`
+            Arithmetic expression to use if opening a complex-valued image.
 
         Returns
         -------
@@ -74,9 +74,7 @@ class Image:
             image_arithmetic = False
         elif complex is True:
             image_arithmetic = "true"
-            if Expression is None:
-                Expression = "AMPLITUDE"
-            file_name = f'{Expression}("{file_name}")'
+            file_name = f'{expression}("{file_name}")'
         image_id = session.call_action(command, directory, file_name, hdu, image_arithmetic, return_path="frameInfo.fileId")
         return cls(session, image_id, file_name)
 
--- a/carta/image.py
+++ b/carta/image.py
@@ -58,7 +58,7 @@ class Image:
         append : boolean
             Whether the image should be appended.
         complex : boolean
-            Whether the image is complex.
+            Whether the image is complex. Set to ``False`` by default.
         expression : a member of :obj:`carta.constants.ArithmeticExpression` or ``None``
             Arithmetic expression to use if opening a complex-valued image. Set to ``None`` if the image is not complex, or to show the default amplitude if the image is complex.
 
--- a/carta/image.py
+++ b/carta/image.py
@@ -42,7 +42,7 @@ class Image:
         self._frame = Macro("", self._base_path)
 
     @classmethod
-    def new(cls, session, path, hdu, append, complex=None):
+    def new(cls, session, path, hdu, append, complex=False, Expression=None):
         """Open or append a new image in the session and return an image object associated with it.
 
         This method should not be used directly. It is wrapped by :obj:`carta.session.Session.open_image` and :obj:`carta.session.Session.append_image`.
@@ -57,7 +57,9 @@ class Image:
             The HDU to open.
         append : boolean
             Whether the image should be appended.
-        complex : a member of :obj:`carta.constants.ArithmeticExpression` or ``None``
+        complex : boolean
+            Whether the image is complex.
+        expression : a member of :obj:`carta.constants.ArithmeticExpression` or ``None``
             Arithmetic expression to use if opening a complex-valued image. Set to ``None`` if the image is not complex, or to show the default amplitude if the image is complex.
 
         Returns
@@ -68,11 +70,13 @@ class Image:
         path = session.resolve_file_path(path)
         directory, file_name = posixpath.split(path)
         command = "appendFile" if append else "openFile"
-        if complex is not None:
-            image_arithmetic = "true"
-            file_name = f'{complex}("{file_name}")'
-        else:
+        if complex is False:
             image_arithmetic = False
+        elif complex is True:
+            image_arithmetic = "true"
+            if Expression is None:
+                Expression = "AMPLITUDE"
+            file_name = f'{Expression}("{file_name}")'
         image_id = session.call_action(command, directory, file_name, hdu, image_arithmetic, return_path="frameInfo.fileId")
         return cls(session, image_id, file_name)
 
--- a/carta/image.py
+++ b/carta/image.py
@@ -42,7 +42,7 @@ class Image:
         self._frame = Macro("", self._base_path)
 
     @classmethod
-    def new(cls, session, path, hdu, append, composite=None):
+    def new(cls, session, path, hdu, append, complex=None):
         """Open or append a new image in the session and return an image object associated with it.
 
         This method should not be used directly. It is wrapped by :obj:`carta.session.Session.open_image` and :obj:`carta.session.Session.append_image`.
@@ -57,6 +57,8 @@ class Image:
             The HDU to open.
         append : boolean
             Whether the image should be appended. By default it is not, and all other open images are closed.
+        complex : None or string
+            Complex value to open a comlex-valued image. The valid options are ``AMPLITUDE``, ``PHASE``, ``REAL`` and ``IMAGINARY``. Default is ``AMPLITUDE``.
 
         Returns
         -------
@@ -65,10 +67,12 @@ class Image:
         """
         path = session.resolve_file_path(path)
         directory, file_name = posixpath.split(path)
-        if composite is not None:
-            file_name = f'{composite}("{file_name}")'
-        image_id = session.call_action("appendFile" if append else "openFile", directory, file_name, hdu, return_path="frameInfo.fileId")
-
+        if complex is not None:
+            file_name = f'{complex}("{file_name}")'
+            image_arithmetic = "true"
+            image_id = session.call_action("appendFile" if append else "openFile", directory, file_name, hdu, image_arithmetic, return_path="frameInfo.fileId")
+        else:
+            image_id = session.call_action("appendFile" if append else "openFile", directory, file_name, hdu, return_path="frameInfo.fileId")
         return cls(session, image_id, file_name)
 
     @classmethod
--- a/carta/image.py
+++ b/carta/image.py
@@ -70,11 +70,11 @@ class Image:
         path = session.resolve_file_path(path)
         directory, file_name = posixpath.split(path)
         command = "appendFile" if append else "openFile"
-        if complex is False:
-            image_arithmetic = False
-        elif complex is True:
-            image_arithmetic = "true"
+        if complex:
+            image_arithmetic = True
             file_name = f'{expression}("{file_name}")'
+        else:
+            image_arithmetic = False
         image_id = session.call_action(command, directory, file_name, hdu, image_arithmetic, return_path="frameInfo.fileId")
         return cls(session, image_id, file_name)
 
--- a/carta/image.py
+++ b/carta/image.py
@@ -42,7 +42,7 @@ class Image:
         self._frame = Macro("", self._base_path)
 
     @classmethod
-    def new(cls, session, path, hdu, append):
+    def new(cls, session, path, hdu, append, composite=None):
         """Open or append a new image in the session and return an image object associated with it.
 
         This method should not be used directly. It is wrapped by :obj:`carta.session.Session.open_image` and :obj:`carta.session.Session.append_image`.
@@ -65,6 +65,8 @@ class Image:
         """
         path = session.resolve_file_path(path)
         directory, file_name = posixpath.split(path)
+        if composite is not None:
+            file_name = f'{composite}("{file_name}")'
         image_id = session.call_action("appendFile" if append else "openFile", directory, file_name, hdu, return_path="frameInfo.fileId")
 
         return cls(session, image_id, file_name)
--- a/carta/image.py
+++ b/carta/image.py
@@ -57,8 +57,8 @@ class Image:
             The HDU to open.
         append : boolean
             Whether the image should be appended. By default it is not, and all other open images are closed.
-        complex : None or string
-            Complex value to open a comlex-valued image. The valid options are ``AMPLITUDE``, ``PHASE``, ``REAL`` and ``IMAGINARY``. Default is ``AMPLITUDE``.
+        complex : a member of :obj:`carta.constants.ArithmeticExpression` or ``None``
+            Arithmetic expression to use if opening a complex-valued image. By default the image is assumed not to be complex.
 
         Returns
         -------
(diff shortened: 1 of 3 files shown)


confluence:
Please turn on logging of scripting request and response messages in the frontend. You will then be able to compare the working and not-working request messages in the console.

When you call the `openFile` action directly, you're adding another positional parameter set to `"true"` -- this is the `imageArithmetic` parameter which I assume makes the expression work. You haven't added that inside `Image.new`.

I-Chenn:
@confluence, @kswang1029 :
I have made a new commit (941b566) and it works with my preliminary test.
The PR is hence ready for review.

I-Chenn:
@confluence @kswang1029 , I will make a modification to deal with the `complex` with a new added constant. Please wait for my next commit.

I-Chenn:
@confluence, @kswang1029:
2358857 is ready for review.

carta/image.py
```diff
@@ -57,6 +57,8 @@ def new(cls, session, path, hdu, append):
             The HDU to open.
         append : boolean
             Whether the image should be appended. By default it is not, and all other open images are closed.
+        complex : None or string
+            Complex value to open a comlex-valued image. The valid options are ``AMPLITUDE``, ``PHASE``, ``REAL`` and ``IMAGINARY``. Default is ``AMPLITUDE``.
```
confluence: These should be enum symbols, not strings.

carta/image.py
```diff
@@ -57,6 +57,8 @@ def new(cls, session, path, hdu, append):
             The HDU to open.
         append : boolean
             Whether the image should be appended. By default it is not, and all other open images are closed.
+        complex : None or string
```
confluence: Should be:

```
a member of :obj:`carta.constants.ComplexValue` or ``None``
```

carta/image.py
confluence: Also, a typo in `complex-valued`. It's not necessary to list all the options here because the user can inspect or look up the enum.

confluence: Also, "complex value" is ambiguous here. Going by the frontend, perhaps "arithmetic expression" is clearer. Also, by default this is None / unset, not amplitude. So perhaps:

"Arithmetic expression to use if opening a complex-valued image. By default the image is assumed not to be complex."

carta/image.py
```diff
@@ -65,8 +67,12 @@ def new(cls, session, path, hdu, append):
...
         path = session.resolve_file_path(path)
         directory, file_name = posixpath.split(path)
-        image_id = session.call_action("appendFile" if append else "openFile", directory, file_name, hdu, return_path="frameInfo.fileId")
-
+        if complex is not None:
+            file_name = f'{complex}("{file_name}")'
+            image_arithmetic = "true"
+            image_id = session.call_action("appendFile" if append else "openFile", directory, file_name, hdu, image_arithmetic, return_path="frameInfo.fileId")
+        else:
+            image_id = session.call_action("appendFile" if append else "openFile", directory, file_name, hdu, return_path="frameInfo.fileId")
```
confluence: I would suggest doing a single call at the end outside the conditionals. Add an `image_arithmetic = False` either before the `complex is not None` conditional or in an `else`, and then you can do the call once at the end (including the `image_arithmetic` parameter).

If you like, maybe move the inline conditional for the action name from the call into its own line -- `command = "appendFile" if append else "openFile"`, and then use `command` in the call.

carta/session.py
```diff
@@ -368,11 +368,13 @@ def open_image(self, path, hdu=""):
             The path to the image file, either relative to the session's current directory or an absolute path relative to the CARTA backend's root directory.
         hdu : {1}
             The HDU to select inside the file.
+        complex : {2}
+            Complex value to open a comlex-valued image. The valid options are ``AMPLITUDE``, ``PHASE``, ``REAL`` and ``IMAGINARY``. Default is ``AMPLITUDE``.
```
confluence: Update this description as well (but you won't need to update the type information, because the validator fills that in -- maybe check if what I suggested in `new` matches the validator's description).

carta/session.py
```diff
@@ -381,8 +383,10 @@ def append_image(self, path, hdu=""):
             The path to the image file, either relative to the session's current directory or an absolute path relative to the CARTA backend's root directory.
         hdu : {1}
             The HDU to select inside the file.
+        complex : {2}
+            Complex value to append a comlex-valued image. The valid options are ``AMPLITUDE``, ``PHASE``, ``REAL`` and ``IMAGINARY``. Default is ``AMPLITUDE``.
```
confluence: Also update this description.

confluence (changes requested):

carta/session.py
I-Chenn: Done in 0b989dc.

carta/session.py
I-Chenn: Done in 0b989dc.

carta/image.py
I-Chenn: Done in 8b8c752.

carta/image.py
I-Chenn: Done in 8b8c752.

carta/session.py
```diff
@@ -368,11 +368,13 @@ def open_image(self, path, hdu=""):
             The path to the image file, either relative to the session's current directory or an absolute path relative to the CARTA backend's root directory.
         hdu : {1}
             The HDU to select inside the file.
+        complex : {2}
+            Arithmetic expression to use if opening a complex-valued image. A member of :obj:`carta.constants.ArithmeticExpression` or ``None``. By default the image is assumed not to be complex.
```
confluence: Don't put the type information in here -- it will be filled in automatically by the validation decorator where you have `{2}`. You can check this by building the docs locally.

carta/session.py
```diff
@@ -381,8 +383,10 @@ def append_image(self, path, hdu=""):
             The path to the image file, either relative to the session's current directory or an absolute path relative to the CARTA backend's root directory.
         hdu : {1}
             The HDU to select inside the file.
+        complex : {2}
+            Arithmetic expression to use if appending a complex-valued image. A member of :obj:`carta.constants.ArithmeticExpression` or ``None``. By default the image is assumed not to be complex.
```
confluence: As above.

carta/image.py
I-Chenn: @confluence, Done in 8b8c752.
As for the default value of `complex`, there are actually two kinds of default. One is that if the image is not a complex image, it is of course opened without any problem. The second is that if the image is a complex image, then we open it without `complex` argument set, the default image that opened is **AMPLITUDE**. I don't know how to describe that in short in the description, do you have any idea about it?

carta/image.py
```diff
@@ -42,7 +42,7 @@ def __init__(self, session, image_id, file_name):
         self._frame = Macro("", self._base_path)
 
     @classmethod
-    def new(cls, session, path, hdu, append):
+    def new(cls, session, path, hdu, append, complex=None):
```
confluence: This is a minor point, but because this method is not intended to be used directly and is always called by `openImage` or `appendImage`, and those methods always pass all of the parameters, the `complex` parameter here should be required, not optional.

And as a result, you should also update the description -- instead of `By default the image is assumed not to be complex.`, put
```
Set to ``None`` if the image is not complex.
```

I actually made a mistake in the description for `append` -- please remove `By default it is not, and all other open images are closed.`, since `append` is required and has no default (how it's set depends on which method calls this method).

(This does *not* apply to `openFile` and `appendFile`; those higher-level functions should keep the existing default and description.)

confluence (changes requested):

carta/image.py
confluence: Oh, you mean that if you try to open a complex image in CARTA using the normal file path, without using an expression, CARTA will automatically detect that it is complex and use amplitude by default?

In that case, the default portion of the description for this parameter in `openFile` and `appendFile` should say:
```
By default, the amplitude will be shown if the image is complex.
```

And in `new`, this portion should say:
```
Set to ``None`` if the image is not complex, or to show the default amplitude if the image is complex.
```

1. Is amplitude a hardcoded default in the frontend, or is it something that the user can configure?
2. What happens if the user uses an arithmetic expression on an image which isn't complex? Will it open normally, or will this cause an error?

I-Chenn: I am not sure about 1., will take a look at it.
As for 2, I have just tested it, it does cause an error. Since the `complex` is applied to `file_name`, it's not able to look for the file name with `complex` attached to it.

carta/session.py
I-Chenn: @confluence, I still have no idea how to use sphinx to build a doc for our python code. I promise I will figure it out this weekend.   Nevertheless, I would like to make sure what does **type information** mean here? Do you mean the sentence  of **A member of :obj:`carta.constants.ArithmeticExpression` or ``None``**?.

kswang1029: make sure the theme is installed first
`pip install sphinx-rtd-theme`
then
`make html` in the docs folder.

kswang1029: this is what it reads in the doc. you may refer to "colormap" docstring in the image class to see how it works as a reference. 

Screen Shot 2023-04-13 at 19 03 07

I-Chenn: Thanks a lot!

I-Chenn:
@confluence, @kswang1029:
I have made several modifications in 25077b2, following the request above. Nevertheless, as we discuss above, there is a derivative issue that the current code is not capable of dealing with the circumstance that **"User specify `complex` argument to a non-complex valued image"**, which may result in an error.
I will try to deal with that today.

confluence:
Right. If that is not allowed, then perhaps the correct way to do this is to have two parameters:

1. `complex`, a boolean, which defaults to false, which determines whether the arithmetic expression is used.
2. `expression`, an arithmetic expression, which defaults to amplitude, and which is ignored if `complex` is false.

This makes it clear to the user that *they* have to tell *us* whether the image is complex, and optionally specify an expression. Otherwise, if we only have one parameter, the user may expect that they can specify a different default arithmetic expression and apply the same parameters to multiple different images, and have the expression apply only to complex images.

I-Chenn:
@confluence, I have added the argument `expression` to `open_image`, `append_image` and `new`. Please take a look at it.

carta/image.py
```diff
@@ -42,7 +42,7 @@ def __init__(self, session, image_id, file_name):
         self._frame = Macro("", self._base_path)
 
     @classmethod
-    def new(cls, session, path, hdu, append):
+    def new(cls, session, path, hdu, append, complex=False, Expression=None):
```
confluence: `expression` should be lowercase everywhere.

confluence: None of these parameters should have defaults here -- they should all be required in this method. Defaults should be set only in the two session methods that call this method.

carta/image.py
```diff
@@ -65,8 +69,15 @@ def new(cls, session, path, hdu, append):
...
         directory, file_name = posixpath.split(path)
-        image_id = session.call_action("appendFile" if append else "openFile", directory, file_name, hdu, return_path="frameInfo.fileId")
-
+        command = "appendFile" if append else "openFile"
+        if complex is False:
+            image_arithmetic = False
+        elif complex is True:
+            image_arithmetic = "true"
+            if Expression is None:
+                Expression = "AMPLITUDE"
```
confluence: This won't be necessary if `expression` defaults to amplitude in the two session methods and is required here.

carta/session.py
```diff
@@ -358,8 +358,8 @@ def set_starting_directory(self, path):
 
     # IMAGES
 
-    @validate(String(), String(r"\d*"))
-    def open_image(self, path, hdu=""):
+    @validate(String(), String(r"\d*"), Boolean(), NoneOr(Constant(ArithmeticExpression)))
+    def open_image(self, path, hdu="", complex=False, expression=None):
```
confluence: Instead of making `expression` default to `None`, please make it default to `ArithmeticExpression.AMPLITUDE` -- this is no longer used to turn arithmetic on and off, so if `complex` is false you just ignore it.

The description should be updated to match:
```
(...) The default is :obj:`carta.constants.ArithmeticExpression.AMPLITUDE`.
```

The validation should be changed to match (no `NoneOr`).

carta/image.py
```diff
@@ -56,7 +56,11 @@ def new(cls, session, path, hdu, append):
         hdu : string
             The HDU to open.
         append : boolean
-            Whether the image should be appended. By default it is not, and all other open images are closed.
+            Whether the image should be appended.
+        complex : boolean
+            Whether the image is complex. Set to ``False`` by default.
+        expression : a member of :obj:`carta.constants.ArithmeticExpression` or ``None``
+            Arithmetic expression to use if opening a complex-valued image. Set to ``None`` if the image is not complex, or to show the default amplitude if the image is complex.
```
confluence: Defaults should be removed.

carta/session.py
```diff
@@ -368,11 +368,15 @@ def open_image(self, path, hdu=""):
...
+        expression : {3}
+            Arithmetic expression to use if opening a complex-valued image. By default, the amplitude will be shown if the image is complex.
         """
-        return Image.new(self, path, hdu, False)
+        return Image.new(self, path, hdu, False, complex, expression)
 
-    @validate(String(), String(r"\d*"))
-    def append_image(self, path, hdu=""):
+    @validate(String(), String(r"\d*"), Boolean(), NoneOr(Constant(ArithmeticExpression)))
+    def append_image(self, path, hdu="", complex=False, expression=None):
```
confluence: As above, the default should be set to amplitude here, and the description and validation should be updated.

confluence (changes requested):
Please also make sure that you've built the documentation locally and checked that these three functions all look right (all the `{...}` parameter types should be filled in, all the `:obj:` references should link to the objects, and all the double backtick literal values should be rendered as monospace).

carta/image.py
I-Chenn: Done in f651be6.

carta/session.py
I-Chenn: Done in f651be6.

carta/image.py
I-Chenn: Done in f651be6

carta/session.py
I-Chenn: Done in f651be6.

carta/image.py
I-Chenn: Done in f651be6.

carta/image.py
```diff
@@ -65,8 +69,13 @@ def new(cls, session, path, hdu, append):
         """
         path = session.resolve_file_path(path)
         directory, file_name = posixpath.split(path)
-        image_id = session.call_action("appendFile" if append else "openFile", directory, file_name, hdu, return_path="frameInfo.fileId")
-
+        command = "appendFile" if append else "openFile"
+        if complex is False:
+            image_arithmetic = False
+        elif complex is True:
+            image_arithmetic = "true"
```
confluence: This should be `True`, not `"true"` (I missed this last time).

confluence (changes requested):
One last thing that I missed -- after that this will be ready for merge.

carta/image.py
confluence: Actually, instead of explicitly comparing to `True` and `False`, this would be more cleanly expressed as:
```
if complex:
    (...)
else:
    (....)
```

I-Chenn: Done in 8e40b99. Thanks a lot.

confluence (approved):

PR #102, opened 2023-04-11
Add step and offset parameters to the Number validator to allow discrete numeric ranges

confluence:
This should make it possible to use the `Number` validator to validate discrete number ranges with a fixed step size and offset. It should not affect existing usage of the validator. Validation doesn't check *type* -- if `1` is a valid value, so is `1.0`.

### DIFF
--- a/carta/image.py
+++ b/carta/image.py
@@ -131,7 +131,7 @@ class Image:
         return self.session.get_value(f"{self._base_path}.{path}")
 
     def macro(self, target, variable):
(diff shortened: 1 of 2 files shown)


confluence:
I have tested this in `set_channel` (which should have integer values, which we can now enforce), and it seems to be working.

(I have merged in the `set_channel` fix from #101 -- I'll merge dev in here after that's merged so that we have a clean diff.)

kswang1029 (approved):

I-Chenn (approved):

PR #108, opened 2023-04-26
Icchen/91 load image via lel

### LINKED ISSUE #91: support loading image via LEL

I-Chenn:
The PR is to resolve #91.

A function called `load_LEL_image` is added to `session.py` to fulfill the issue.

User need to provide the **_directory_** (instead of _path_, will explain later) of the image(s) being opened and the LEL **_expression_** to call `Image.new`.

In `Image.new`, the `expression` that was used for **complex-valued image** is now sharing its usage with **LEL arithmetic expression**. Since user only provide `directory` in `load_LEL_image`, hence the `file_name` must be an **_empty string_** after the processes of `resolve_file_path` and `posixpath.split`. Thus I am able to add a condition:
```
elif complex is False and file_name == "":
    image_arithmetic = True
    file_name = expression
```
to apply the LEL arithmetic expression. (A little bit tricky, but it does work...)

Furthermore, I renamed the `ArithmeticExpression` to `ComplexExpression` in `constants.py` to better clarify the difference between the **_LEL arithmetic expression_** and the **_complex expression_**.

Check for the result:
https://drive.google.com/file/d/1t0CRUDrSCBDWHPWTalCEp9PHlZAIQfe8/view?usp=sharing

@confluence, @kswang1029:  Please feel free to let me know if you have a better idea.

### DIFF
--- a/carta/image.py
+++ b/carta/image.py
@@ -6,7 +6,7 @@ import posixpath
 
 from .constants import Colormap, Scaling, SmoothingMode, ContourDashMode, Polarization
 from .util import Macro, cached
-from .validation import validate, Number, Color, Constant, Boolean, NoneOr, IterableOf, Evaluate, Attr, Attrs, OneOf
+from .validation import validate, Number, Color, Constant, Boolean, NoneOr, IterableOf, Evaluate, Attr, Attrs, OneOf, String
 
 
 class Image:
@@ -42,25 +42,26 @@ class Image:
         self._frame = Macro("", self._base_path)
 
     @classmethod
+    @validate(String(), String(), String(), Boolean(), Boolean())
     def new(cls, session, directory, file_name, hdu, append, image_arithmetic):
         """Open or append a new image in the session and return an image object associated with it.
 
-        This method should not be used directly. It is wrapped by :obj:`carta.session.Session.open_image`, :obj:`carta.session.Session.open_complex_image` and :obj:`carta.session.Session.open_complex_image`.
+        This method should not be used directly. It is wrapped by :obj:`carta.session.Session.open_image`, :obj:`carta.session.Session.open_complex_image` and :obj:`carta.session.Session.open_LEL_image`.
 
         Parameters
         ----------
         session : :obj:`carta.session.Session`
             The session object.
-        directory : string
-            The directory to the image file, either relative to the session's current directory or an absolute path relative to the CARTA backend's root directory.
-        file_name : string
-            The name of the image file or the LEL expression for complex image and arithmeic image.
-        hdu : string
+        directory : {0}
+            The directory containing the image file, or the base directory for the LEL arithmetic expression, as an absolute path relative to the CARTA backend's root directory.
+        file_name : {1}
+            The name of the image file, or a LEL arithmetic expression.
+        hdu : {2}
             The HDU to open.
-        append : boolean
+        append : {3}
             Whether the image should be appended.
-        image_arihmetic : boolean
-            Whether the LEL expression is used.
+        image_arihmetic : {4}
+            Whether the file name should be interpreted as a LEL expression.
 
         Returns
         -------
--- a/carta/image.py
+++ b/carta/image.py
@@ -42,7 +42,7 @@ class Image:
         self._frame = Macro("", self._base_path)
 
     @classmethod
-    def new(cls, session, directory, file_name, hdu, append, image_arithmetic, make_active=True, update_directory=False):
+    def new(cls, session, path, hdu, append, complex, image_arithmetic, expression, make_active=True, update_directory=False):
         """Open or append a new image in the session and return an image object associated with it.
 
         This method should not be used directly. It is wrapped by :obj:`carta.session.Session.open_image`, :obj:`carta.session.Session.open_complex_image` and :obj:`carta.session.Session.open_LEL_image`.
@@ -51,16 +51,18 @@ class Image:
         ----------
         session : :obj:`carta.session.Session`
             The session object.
-        directory : string
-            The directory containing the image file, or the base directory for the LEL arithmetic expression, as an absolute path relative to the CARTA backend's root directory.
-        file_name : string
-            The name of the image file, or a LEL arithmetic expression.
+        path : string
+            The path to the image file, either relative to the session's current directory or an absolute path relative to the CARTA backend's root directory.
         hdu : string
             The HDU to open.
         append : boolean
             Whether the image should be appended.
+        complex : boolean
+            Whether the image is complex.
         image_arihmetic : boolean
-            Whether the file name should be interpreted as a LEL expression. Ignored if the image is not complex.
+            Whether the file name should be interpreted as a complex expression or a LEL expression.
+        expression : string
+            The name of the image file, or a LEL arithmetic expression.
         make_active : boolean
             Whether the image should be made active in the frontend. This only applies if an image is being appended. The default is ``True``.
         update_directory : boolean
@@ -72,6 +74,14 @@ class Image:
             A new image object.
         """
         command = "appendFile" if append else "openFile"
+
+        path = session.resolve_file_path(path)
+        directory, file_name = posixpath.split(path)
+
+        if complex and image_arithmetic:
+            file_name = f'{expression}("{file_name}")'
+        elif not complex and image_arithmetic:
+            file_name = expression
 
         params = [directory, file_name, hdu, image_arithmetic]
         if append:
--- a/carta/image.py
+++ b/carta/image.py
@@ -42,42 +42,32 @@ class Image:
         self._frame = Macro("", self._base_path)
 
     @classmethod
-    def new(cls, session, path, hdu, append, complex, expression):
+    def new(cls, session, directory, file_name, hdu, append, image_arithmetic):
         """Open or append a new image in the session and return an image object associated with it.
 
-        This method should not be used directly. It is wrapped by :obj:`carta.session.Session.open_image` and :obj:`carta.session.Session.append_image`.
+        This method should not be used directly. It is wrapped by :obj:`carta.session.Session.open_image`, :obj:`carta.session.Session.open_complex_image` and :obj:`carta.session.Session.open_complex_image`.
 
         Parameters
         ----------
         session : :obj:`carta.session.Session`
             The session object.
-        path : string
-            The path to the image file, either relative to the session's current directory or an absolute path relative to the CARTA backend's root directory.
+        directory : string
+            The directory to the image file, either relative to the session's current directory or an absolute path relative to the CARTA backend's root directory.
+        file_name : string
+            The name of the image file or the LEL expression for complex image and arithmeic image.
         hdu : string
             The HDU to open.
         append : boolean
             Whether the image should be appended.
-        complex : boolean
-            Whether the image is complex.
-        expression : string or a member of :obj:`carta.constants.ComplexExpression`
-            Arithmetic expression or complex expression to use if opening a complex-valued image.
+        image_arihmetic : boolean
+            Whether the LEL expression is used.
 
         Returns
         -------
         :obj:`carta.image.Image`
             A new image object.
         """
-        path = session.resolve_file_path(path)
-        directory, file_name = posixpath.split(path)
         command = "appendFile" if append else "openFile"
-        if complex:
-            image_arithmetic = True
-            file_name = f'{expression}("{file_name}")'
-        elif complex is False and file_name == "":
-            image_arithmetic = True
-            file_name = expression
-        else:
-            image_arithmetic = False
         image_id = session.call_action(command, directory, file_name, hdu, image_arithmetic, return_path="frameInfo.fileId")
         return cls(session, image_id, file_name)
 
--- a/carta/image.py
+++ b/carta/image.py
@@ -52,7 +52,7 @@ class Image:
         directory : string
             The directory containing the image file or the base directory for the LEL arithmetic expression, either relative to the session's current directory or an absolute path relative to the CARTA backend's root directory
         file_name : string
-            The name of the image file, or a LEL arithmetic expression.
+            The name of the image file, complex expression or a LEL arithmetic expression.
         hdu : string
             The HDU to open.
         append : boolean
--- a/carta/image.py
+++ b/carta/image.py
@@ -59,7 +59,7 @@ class Image:
             The HDU to open.
         append : boolean
             Whether the image should be appended.
-        image_arihmetic : boolean
+        image_arithmetic : boolean
             Whether the file name should be interpreted as a LEL expression. Ignored if the image is not complex.
         make_active : boolean
             Whether the image should be made active in the frontend. This only applies if an image is being appended. The default is ``True``.
@@ -72,6 +72,7 @@ class Image:
             A new image object.
         """
         command = "appendFile" if append else "openFile"
+        directory = session.resolve_file_path(directory)
 
         params = [directory, file_name, hdu, image_arithmetic]
         if append:
--- a/carta/image.py
+++ b/carta/image.py
@@ -50,7 +50,7 @@ class Image:
         session : :obj:`carta.session.Session`
             The session object.
         directory : string
-            The directory containing the image file, or the base directory for the LEL arithmetic expression, as an absolute path relative to the CARTA backend's root directory.
+            The directory containing the image file or the base directory for the LEL arithmetic expression, either relative to the session's current directory or an absolute path relative to the CARTA backend's root directory
         file_name : string
             The name of the image file, or a LEL arithmetic expression.
         hdu : string
@@ -58,7 +58,7 @@ class Image:
         append : boolean
             Whether the image should be appended.
         image_arithmetic : boolean
-            Whether the file name should be interpreted as a LEL expression. Ignored if the image is not complex.
+            Whether the file name should be interpreted as a complex expression or a LEL expression.
         make_active : boolean
             Whether the image should be made active in the frontend. This only applies if an image is being appended. The default is ``True``.
         update_directory : boolean
--- a/carta/image.py
+++ b/carta/image.py
@@ -42,7 +42,6 @@ class Image:
         self._frame = Macro("", self._base_path)
 
     @classmethod
-    @validate(String(), String(), String(), Boolean(), Boolean())
     def new(cls, session, directory, file_name, hdu, append, image_arithmetic):
         """Open or append a new image in the session and return an image object associated with it.
 
@@ -52,15 +51,15 @@ class Image:
         ----------
         session : :obj:`carta.session.Session`
             The session object.
-        directory : {0}
+        directory : string
             The directory containing the image file, or the base directory for the LEL arithmetic expression, as an absolute path relative to the CARTA backend's root directory.
-        file_name : {1}
+        file_name : string
             The name of the image file, or a LEL arithmetic expression.
-        hdu : {2}
+        hdu : string
             The HDU to open.
-        append : {3}
+        append : boolean
             Whether the image should be appended.
-        image_arihmetic : {4}
+        image_arihmetic : boolean
             Whether the file name should be interpreted as a LEL expression.
 
         Returns
--- a/carta/image.py
+++ b/carta/image.py
@@ -2,11 +2,9 @@
 
 Image objects should not be instantiated directly, and should only be created through methods on the :obj:`carta.session.Session` object.
 """
-import posixpath
-
 from .constants import Colormap, Scaling, SmoothingMode, ContourDashMode, Polarization, CoordinateSystem, SpatialAxis
 from .util import Macro, cached, PixelValue, AngularSize, WorldCoordinate
-from .validation import validate, Number, Color, Constant, Boolean, NoneOr, IterableOf, Evaluate, Attr, Attrs, OneOf, Size, Coordinate, String
+from .validation import validate, Number, Color, Constant, Boolean, NoneOr, IterableOf, Evaluate, Attr, Attrs, OneOf, Size, Coordinate
 
 
 class Image:
--- a/carta/image.py
+++ b/carta/image.py
@@ -42,7 +42,7 @@ class Image:
         self._frame = Macro("", self._base_path)
 
     @classmethod
-    def new(cls, session, path, hdu, append, complex, image_arithmetic, expression, make_active=True, update_directory=False):
+    def new(cls, session, directory, file_name, hdu, append, image_arithmetic, make_active=True, update_directory=False):
         """Open or append a new image in the session and return an image object associated with it.
 
         This method should not be used directly. It is wrapped by :obj:`carta.session.Session.open_image`, :obj:`carta.session.Session.open_complex_image` and :obj:`carta.session.Session.open_LEL_image`.
@@ -51,18 +51,16 @@ class Image:
         ----------
         session : :obj:`carta.session.Session`
             The session object.
-        path : string
-            The path to the image file, either relative to the session's current directory or an absolute path relative to the CARTA backend's root directory.
+        directory : string
+            The directory containing the image file, or the base directory for the LEL arithmetic expression, as an absolute path relative to the CARTA backend's root directory.
+        file_name : string
+            The name of the image file, or a LEL arithmetic expression.
         hdu : string
             The HDU to open.
         append : boolean
             Whether the image should be appended.
-        complex : boolean
-            Whether the image is complex.
         image_arihmetic : boolean
-            Whether the file name should be interpreted as a complex expression or a LEL expression.
-        expression : string
-            The name of the image file, or a LEL arithmetic expression.
+            Whether the file name should be interpreted as a LEL expression. Ignored if the image is not complex.
         make_active : boolean
             Whether the image should be made active in the frontend. This only applies if an image is being appended. The default is ``True``.
         update_directory : boolean
@@ -74,14 +72,6 @@ class Image:
             A new image object.
         """
         command = "appendFile" if append else "openFile"
-
-        path = session.resolve_file_path(path)
-        directory, file_name = posixpath.split(path)
-
-        if complex and image_arithmetic:
-            file_name = f'{expression}("{file_name}")'
-        elif not complex and image_arithmetic:
-            file_name = expression
 
         params = [directory, file_name, hdu, image_arithmetic]
         if append:
--- a/carta/image.py
+++ b/carta/image.py
@@ -52,13 +52,13 @@ class Image:
         directory : string
             The directory containing the image file or the base directory for the LEL arithmetic expression, either relative to the session's current directory or an absolute path relative to the CARTA backend's root directory
         file_name : string
-            The name of the image file, complex expression or a LEL arithmetic expression.
+            The name of the image file, or a LEL arithmetic expression.
         hdu : string
             The HDU to open.
         append : boolean
             Whether the image should be appended.
         image_arithmetic : boolean
-            Whether the file name should be interpreted as a complex expression or a LEL expression.
+            Whether the file name should be interpreted as a LEL expression.
         make_active : boolean
             Whether the image should be made active in the frontend. This only applies if an image is being appended. The default is ``True``.
         update_directory : boolean
--- a/carta/image.py
+++ b/carta/image.py
@@ -50,7 +50,7 @@ class Image:
         session : :obj:`carta.session.Session`
             The session object.
         directory : string
-            The directory containing the image file or the base directory for the LEL arithmetic expression, either relative to the session's current directory or an absolute path relative to the CARTA backend's root directory
+            The directory containing the image file or the base directory for the LEL arithmetic expression, either relative to the session's current directory or an absolute path relative to the CARTA backend's root directory.
         file_name : string
             The name of the image file, or a LEL arithmetic expression.
         hdu : string
@@ -62,7 +62,7 @@ class Image:
         make_active : boolean
             Whether the image should be made active in the frontend. This only applies if an image is being appended. The default is ``True``.
         update_directory : boolean
-            Whether the starting directory of the frontend file browser should be updated to the parent directory of the image. The default is ``False``.
+            Whether the starting directory of the frontend file browser should be updated to the directory provided. The default is ``False``.
 
         Returns
         -------
--- a/carta/image.py
+++ b/carta/image.py
@@ -59,8 +59,8 @@ class Image:
             Whether the image should be appended.
         complex : boolean
             Whether the image is complex.
-        expression : a member of :obj:`carta.constants.ArithmeticExpression`
-            Arithmetic expression to use if opening a complex-valued image.
+        expression : string or a member of :obj:`carta.constants.ComplexExpression`
+            Arithmetic expression or complex expression to use if opening a complex-valued image.
 
         Returns
         -------
@@ -73,6 +73,9 @@ class Image:
         if complex:
             image_arithmetic = True
             file_name = f'{expression}("{file_name}")'
+        elif complex is False and file_name == "":
+            image_arithmetic = True
+            file_name = expression
         else:
             image_arithmetic = False
         image_id = session.call_action(command, directory, file_name, hdu, image_arithmetic, return_path="frameInfo.fileId")
--- a/carta/session.py
+++ b/carta/session.py
@@ -370,7 +370,7 @@ class Session:
         hdu : {1}
             The HDU to select inside the file.
         append : {2}
-            Whether the image should be appended. Default is ``False``.
+            Whether the image should be appended to existing images. By default this is ``False`` and any existing open images are closed.
         """
         path = self.resolve_file_path(path)
         directory, file_name = posixpath.split(path)
@@ -385,9 +385,9 @@ class Session:
         path : {0}
             The path to the complex-valued image file, either relative to the session's current directory or an absolute path relative to the CARTA backend's root directory.
         expression : {1}
-            The complex expression for opening a complex-valued image.
+            The complex expression for opening a complex-valued image. The default is :obj:`carta.constants.ComplexExpression.AMPLITUDE`.
         append : {2}
-            Whether the image should be appended. Default is ``False``.
+            Whether the image should be appended to existing images. By default this is ``False`` and any existing open images are closed.
         """
         path = self.resolve_file_path(path)
         directory, file_name = posixpath.split(path)
@@ -405,7 +405,7 @@ class Session:
         directory : {1}
             The base directory for the LEL expression, either relative to the session's current directory or an absolute path relative to the CARTA backend's root directory. Defaults to the session's current directory.
         append : {2}
-            Whether the image should be appended. Default is ``False``.
+            Whether the image should be appended to existing images. By default this is ``False`` and any existing open images are closed.
         """
         if directory is None:
             directory = self.pwd()
--- a/carta/session.py
+++ b/carta/session.py
@@ -355,7 +355,6 @@ class Session:
         update_directory : {3}
             Whether the starting directory of the frontend file browser should be updated to the parent directory of the image. The default is ``False``.
         """
-        path = self.resolve_file_path(path)
         directory, file_name = posixpath.split(path)
         return Image.new(self, directory, file_name, hdu, append, False, update_directory=update_directory)
 
@@ -374,7 +373,6 @@ class Session:
         update_directory : {3}
             Whether the starting directory of the frontend file browser should be updated to the parent directory of the image. The default is ``False``.
         """
-        path = self.resolve_file_path(path)
         directory, file_name = posixpath.split(path)
         expression = f'{expression}("{file_name}")'
         return Image.new(self, directory, expression, "", append, True, update_directory=update_directory)
@@ -394,7 +392,6 @@ class Session:
         update_directory : {3}
             Whether the starting directory of the frontend file browser should be updated to the parent directory of the image. The default is ``False``.
         """
-        directory = self.resolve_file_path(directory)
         return Image.new(self, directory, expression, "", append, True, update_directory=update_directory)
 
     def image_list(self):
--- a/carta/session.py
+++ b/carta/session.py
@@ -393,11 +393,11 @@ class Session:
         """
         path = self.resolve_file_path(path)
         directory, file_name = posixpath.split(path)
-        file_name = f'{expression}("{file_name}")'
-        return Image.new(self, directory, file_name, hdu, append, True)
+        expression = f'{expression}("{file_name}")'
+        return Image.new(self, directory, expression, hdu, append, True)
 
     @validate(String(), String(), String(r"\d*"), Boolean())
-    def open_LEL_image(self, expression, directory, hdu="", append=False):
+    def open_LEL_image(self, expression, directory=None, hdu="", append=False):
         """Open or append a new image via the Lattice Expression Language (LEL) interface.
 
         Parameters
@@ -405,16 +405,16 @@ class Session:
         expression : {0}
             The LEL arithmetic expression.
         directory : {1}
-            The directory to the image file(s), either relative to the session's current directory or an absolute path relative to the CARTA backend's root directory.
+            The base directory for the LEL expression, either relative to the session's current directory or an absolute path relative to the CARTA backend's root directory. Defaults to the session's current directory.
         hdu : {2}
             The HDU to select inside the file.
         append : {3}
             Whether the image should be appended. Default is ``False``.
         """
-        file_name = expression
-        if "/" in expression:
+        if directory is None:
             directory = self.pwd()
-        return Image.new(self, directory, file_name, hdu, append, True)
+        directory = self.resolve_file_path(directory)
+        return Image.new(self, directory, expression, hdu, append, True)
 
     def image_list(self):
         """Return the list of currently open images.
--- a/carta/session.py
+++ b/carta/session.py
@@ -10,7 +10,7 @@ import base64
 import posixpath
 
 from .image import Image
-from .constants import CoordinateSystem, LabelType, BeamType, PaletteColor, Overlay, PanelMode, GridMode, ComplexExpression, NumberFormat
+from .constants import CoordinateSystem, LabelType, BeamType, PaletteColor, Overlay, PanelMode, GridMode, ComplexComponent, NumberFormat
 from .backend import Backend
 from .protocol import Protocol
 from .util import logger, Macro, split_action_path, CartaBadID, CartaBadSession, CartaBadUrl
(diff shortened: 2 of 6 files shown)


kswang1029:
Can the current implementation support forming an LEL image with two images at different directories? (ie two images with different absolute paths)

kswang1029:
We are using LEL to handle how a complex image is loaded in CARTA, not vice verse.

I-Chenn:
> Can the current implementation support forming an LEL image with two images at different directories? (ie two images with different absolute paths)

No, with the current implementation, the LEL image can only be formed with the images from the same directory, which is the same scenario with the front-end UI. Do we need to make the process with images from different directories?

kswang1029:
> > Can the current implementation support forming an LEL image with two images at different directories? (ie two images with different absolute paths)
> 
> No, with the current implementation, the LEL image can only be formed with the images from the same directory, which is the same scenario with the front-end UI. Do we need to make the process with images from different directories?

The current GUI does support the case. You can give it a try.

carta/session.py
```diff
@@ -388,10 +388,28 @@ def append_image(self, path, hdu="", complex=False, expression=ArithmeticExpress
...
+        """Open or append a new image via the Lattice Expression Language (LEL) interface.
+
+        Parameters
+        ----------
+        directory : {0}
+            The directory of the image file(s), either relative to the session's current directory or an absolute path relative to the CARTA backend's root directory.
+        expression : {1}
+            The LEL arithmetic expression.
+        hdu : {2}
+            The HDU to select inside the file.
```
confluence: How is this value used in the frontend if a LEL expression with multiple images is provided? Is the same HDU applied to each image? Is it not applied at all?

carta/image.py
```diff
@@ -73,6 +73,9 @@ def new(cls, session, path, hdu, append, complex, expression):
         if complex:
             image_arithmetic = True
             file_name = f'{expression}("{file_name}")'
+        elif complex is False and file_name == "":
```
confluence: I would prefer not to guess here, and to have explicit flags set by the functions in the session. I will make a more detailed suggestion in a comment for unifying the complex expression and arbitrary LEL expressions.

confluence (changes requested):

carta/session.py
```diff
@@ -388,10 +388,28 @@ def append_image(self, path, hdu="", complex=False, expression=ArithmeticExpress
...
         """
         return Image.new(self, path, hdu, True, complex, expression)
 
+    @validate(String(), String(), String(r"\d*"), Boolean())
+    def load_LEL_image(self, directory, expression, hdu="", append=False):
+        """Open or append a new image via the Lattice Expression Language (LEL) interface.
+
+        Parameters
+        ----------
+        directory : {0}
```
kswang1029: there is a pwd() function in Session. So I am not sure it is really necessary to have "directory" as a user input.

confluence: Can full paths be included in the LEL expression? If they are, is this directory even used? Can we make this optional, and default to the file browser starting directory, like we do in the Stokes hypercube function?

carta/session.py
```diff
@@ -388,10 +388,28 @@ def append_image(self, path, hdu="", complex=False, expression=ArithmeticExpress
         complex : {2}
             Whether the image is complex. Set to ``False`` by default.
         expression : {3}
-            Arithmetic expression to use if appending a complex-valued image. The default is :obj:`carta.constants.ArithmeticExpression.AMPLITUDE`.
+            Complex expression to use if appending a complex-valued image. The default is :obj:`carta.constants.ComplexExpression.AMPLITUDE`.
         """
         return Image.new(self, path, hdu, True, complex, expression)
 
+    @validate(String(), String(), String(r"\d*"), Boolean())
+    def load_LEL_image(self, directory, expression, hdu="", append=False):
```
confluence: I think this should be called `open_LEL_image`, for consistency. Possibly we should rename the hypercube function to `open_stokes_hypercube` as well.

carta/session.py
kswang1029: at the frontend we can do
`"/Users/kswang/set_QA_e2e_v2/dice_one.fits" + "/Users/kswang/carta_image_pool/set_dice/dice_four.fits"`
Screen Shot 2023-04-26 at 18 26 27

carta/session.py
I-Chenn: But the form `open_something` does not include the append function. The `load_stokes_hypercube` and this `load_LEL_image` allows users to set `append` to `True`. That was my understanding, or I just overthought about that?

confluence:
Here's my suggestion for revising / unifying all these functions:

### Image.new

We make the `Image.new`parameters closely mirror those of `openImage` and `appendImage`. That means that we remove from this function all of the logic for guessing whether the image is complex or a LEL expression. We also remove the path splitting logic from here, since it's not user-facing, and a single path parameter doesn't make sense for some of the high-level functions. The only logic in here should be determining whether to append or not:
```
def new(cls, session, directory, file_name, hdu, append, image_arithmetic):
    command = "appendFile" if append else "openFile"
    image_id = session.call_action(command, directory, file_name, hdu, image_arithmetic, return_path="frameInfo.fileId")
```

### Session.open_image and Session.append_image:

We remove `append_image` and add an `append` flag to `open_image`. This extra shortcut method used to make sense, but now we'd either have it apply only to ordinary files *or* have to add one for each of the new file-opening functions. So it's better to remove it and standardise on the flag.

We keep the single `path` parameter and do the splitting in this function.

We remove the complex functionality from `open_image` and put it in a new function called `open_complex_image`.

### Session.open_complex_image:

In this function we no longer need a `complex` flag, because the use of this function is the flag. We keep the single `path` parameter and do the splitting here as well. We construct the expression and put it in the `file_name` parameter we pass to `new`, and set `image_arithmetic` to true as well.

### Session.open_LEL_image:

From this function we call `new` with the given directory (which should probably be optional, and default to the browser starting directory), the expression as `file_name`, and `image_arithmetic` turned on.

So in the end we will have (in addition to the hypercube function) three separate functions for opening a plain image, a complex image, and a LEL expression, each of which can append. The logic which is specific to these use cases is performed in those functions, and the underlying `new` function is very simple and doesn't try to guess which mode is being used.

carta/session.py
confluence: This is addressed in the longer comment I just made -- I think we need to retire `append_image` and add an `append` flag to `open_image`.

I-Chenn:
@confluence @kswang1029 :
The commit 33bccc1 completes part of the [requested modification](https://github.com/CARTAvis/carta-python/pull/108#issuecomment-1523190798) (`open_image`, `open_complex_image`) . Nevertheless, the `open_LEL_image` is still in need of some configuration to the default value of `directory` argument. Will do that later.

carta/session.py
```diff
@@ -358,39 +359,62 @@ def set_starting_directory(self, path):
...
-            Whether the image is complex. Set to ``False`` by default.
-        expression : {3}
-            Arithmetic expression to use if appending a complex-valued image. The default is :obj:`carta.constants.ArithmeticExpression.AMPLITUDE`.
+        append : {3}
+            Whether the image should be appended. Default is ``False``.
         """
-        return Image.new(self, path, hdu, True, complex, expression)
+        file_name = expression
+        if "/" in expression:
+            directory = self.pwd()
```
confluence: While it may be a good idea to default to the wrapper session's `pwd` rather than the GUI browser's current directory, what is the logic for this conditional? What are you trying to check for by checking if the expression contains `/`? It could be anywhere in the string, and doesn't mean that any or all of the paths are absolute. And I don't think that we should be overriding the directory that the user provides. It makes more sense for this to be the default.

If we do use `pwd` here instead of the current GUI browser directory, we should do the same in the stokes hypercube function.

confluence (changes requested):

carta/session.py
I-Chenn: The reason I want to check if the `expression` contains `/` is that there are two types of `LEL` usage. The first one is that user provides path (both **_directories_** and **_files_**) in `expression`, like:
```
expression = '"/media/icchen/disk/fits/Orion_source_I_Stokes_I.fits"+"/media/icchen/disk/fits/Orion_source_I_Stokes_Q.fits"'
``` 
In this case, the `directory` argument doesn't really matter and we can adopt either `pwd` or `Macro("fileBrowserStore", "startingDirectory")` to make the function work.

Nevertheless, the other one is that user only provide `file_name` in `expression`, like:
```
expression = '"Orion_source_I_Stokes_I.fits"+"Orion_source_I_Stokes_Q.fits"'
```
In this case, the `directory` argument **_does matter_** and we do need this argument to locate the files to open. If the directory is not provided correctly in this case, the execution of the wrapper even crashes the opening CARTA session down.

Therefore, I was trying to make different conditions for these two circumstances by checking if `expression` contains `/`, but not yet able to figure out a promising idea for the second circumstance since we still, regardless, need that `directory` argument.

confluence: But those are not the only two possibilities -- surely some of the paths could be relative and some could be absolute. And this check doesn't tell you anything -- if the whole expression contains a `/` it could be anywhere. It could be in the middle of a relative path, like `subdirectory/filename.fits`. 

You shouldn't be guessing whether the directory is required -- if it's not required it won't be used anyway. If the user provides a directory, you should pass in that directory. But make it optional, and set it to one of the default directories if it's `None`.

If you're going to use `pwd` (which is a good idea), make sure to change this in the hypercube function as well (it's currently using the GUI browser's starting directory, which may not be the same). The documentation for this parameter should be something like `The base directory for the LEL expression, either relative to the session's current directory or an absolute path relative to the CARTA backend's root directory. Defaults to the session's current directory.`

confluence: Sorry; that last part is a little misleading. We have established that in the hypercube function that directory is never actually used (because the individual images have their own separate directories). So changing it should have no effect. But I would like to remove the code which sets it to the GUI browser's starting directory, because we should never be defaulting to that in the wrapper code (we should use the session's `pwd` instead), and leaving it in may be confusing later.

I have remembered that we should be normalizing the LEL function's directory parameter (which may be used), like we do in the other functions: `directory = self.resolve_file_path(directory)`. We do this to convert all relative paths to absolute paths using `pwd` as the base (that's how we can make everything relative to our `pwd` and bypass the GUI browser's current directory, which changes when we open a file).

I-Chenn:
@confluence:
I was meant to make the commit after figuring out how to apply hdu for `LEL`, but then I decided to make this f97b83d first to at least settle the `directory` issue down.
The modification in f97b83d follows your suggestion and should make the function work for most of the cases, except for the circumstance that user only provide file name for `expression` and doesn't provide the `directory`. An error may occur in this circumstance.
If this is OK, I will also modify the default `directory` to `pwd` for `hypercube` case in its branch.
I will also work on how to deal with hdu later.

confluence:
I just looked at the frontend function, and the HDU parameter is set to the default `""` (i.e. completely ignored) if image arithmetic is enabled. That means that it should be removed from the input parameters for both the LEL image function *and* the complex image function, and just set to `""` inside these functions.

carta/image.py
```diff
@@ -42,39 +42,32 @@ def __init__(self, session, image_id, file_name):
...
+        This method should not be used directly. It is wrapped by :obj:`carta.session.Session.open_image`, :obj:`carta.session.Session.open_complex_image` and :obj:`carta.session.Session.open_complex_image`.
 
         Parameters
         ----------
         session : :obj:`carta.session.Session`
             The session object.
-        path : string
-            The path to the image file, either relative to the session's current directory or an absolute path relative to the CARTA backend's root directory.
+        directory : string
+            The directory to the image file, either relative to the session's current directory or an absolute path relative to the CARTA backend's root directory.
```
confluence: This description should read: `The directory containing the image file, or the base directory for the LEL arithmetic expression, as an absolute path relative to the CARTA backend's root directory.`

(The directories which are passed into this function have already been normalized by the session functions, so relative paths have been converted to absolute paths at this point.)

carta/image.py
```diff
@@ -42,39 +42,32 @@ def __init__(self, session, image_id, file_name):
...
+        file_name : string
+            The name of the image file or the LEL expression for complex image and arithmeic image.
```
confluence: This description should read: `The name of the image file, or a LEL arithmetic expression.`

carta/image.py
```diff
@@ -42,39 +42,32 @@ def __init__(self, session, image_id, file_name):
...
         hdu : string
             The HDU to open.
         append : boolean
             Whether the image should be appended.
-        complex : boolean
-            Whether the image is complex.
-        expression : a member of :obj:`carta.constants.ArithmeticExpression`
-            Arithmetic expression to use if opening a complex-valued image.
+        image_arihmetic : boolean
+            Whether the LEL expression is used.
```
confluence: `Whether the file name should be interpreted as a LEL expression.`

carta/image.py
```diff
@@ -42,39 +42,32 @@ def __init__(self, session, image_id, file_name):
         self._frame = Macro("", self._base_path)
 
     @classmethod
-    def new(cls, session, path, hdu, append, complex, expression):
+    def new(cls, session, directory, file_name, hdu, append, image_arithmetic):
```
confluence: When I wrote this I thought that I couldn't use the validation decorator here, but I think that it is actually possible. Please try adding the validation decorator (put it between the `@classmethod` decorator and the function definition). This includes replacing the handwritten types with `{0}` and so on.

carta/session.py
```diff
@@ -358,39 +359,58 @@ def set_starting_directory(self, path):
...
         path : {0}
             The path to the image file, either relative to the session's current directory or an absolute path relative to the CARTA backend's root directory.
         hdu : {1}
             The HDU to select inside the file.
-        complex : {2}
-            Whether the image is complex. Set to ``False`` by default.
-        expression : {3}
-            Arithmetic expression to use if opening a complex-valued image. The default is :obj:`carta.constants.ArithmeticExpression.AMPLITUDE`.
+        append : {2}
+            Whether the image should be appended. Default is ``False``.
```
confluence: This description should read:
```
Whether the image should be appended to existing images. By default this is ``False`` and any existing open images are closed.
```

(The same change should be applied to the other functions.)

carta/session.py
```diff
@@ -358,39 +359,58 @@ def set_starting_directory(self, path):
...
-            The HDU to select inside the file.
-        complex : {2}
-            Whether the image is complex. Set to ``False`` by default.
-        expression : {3}
-            Arithmetic expression to use if appending a complex-valued image. The default is :obj:`carta.constants.ArithmeticExpression.AMPLITUDE`.
-        """
-        return Image.new(self, path, hdu, True, complex, expression)
+            The path to the complex-valued image file, either relative to the session's current directory or an absolute path relative to the CARTA backend's root directory.
+        expression : {1}
+            The complex expression for opening a complex-valued image.
```
confluence: Please add back:
```
The default is :obj:`carta.constants.ArithmeticExpression.AMPLITUDE`.
```

confluence (changes requested):

carta/session.py
I-Chenn: Done in c0838d8, however, I have modified `ArithmeticExpression` to `ComplexExpression`.

carta/session.py
I-Chenn: Done in c0838d8.

carta/image.py
I-Chenn: Done in c0838d8.
Nevertheless, I just added validators for `directory`, `file_name`, `hdu`, `append` and `image_arithmetic`. Do I need to add validator for `session`?

carta/image.py
I-Chenn: Done in c0838d8.

carta/image.py
I-Chenn: Done in c0838d8.

carta/image.py
I-Chenn: Done in c0838d8.

carta/image.py
confluence: Yes, the validation decorator parameters have to match the function parameters, or everything will be shifted up by one and nothing will work. You should be able to use `InstanceOf` and the session class.

I think I initially decided not to use validation here because the function isn't meant to be used directly, but it's exposed and it could be, so it's probably a good idea. I haven't previously tested chaining the decorators like this, so please be sure to test calling the function directly to make sure that everything is working.

carta/session.py
confluence: Sorry, one more thing: the first part should read: `The complex expression to use when opening the image.` (The old description came from the more generic function; this function is *only* for complex-valued images.)

carta/session.py
```diff
@@ -358,39 +359,58 @@ def set_starting_directory(self, path):
...
+        append : {2}
+            Whether the image should be appended to existing images. By default this is ``False`` and any existing open images are closed.
+        """
+        path = self.resolve_file_path(path)
+        directory, file_name = posixpath.split(path)
+        expression = f'{expression}("{file_name}")'
+        return Image.new(self, directory, expression, "", append, True)
+
+    @validate(String(), String(), Boolean())
+    def open_LEL_image(self, expression, directory=None, append=False):
```
kswang1029: Why directory is necessary here? What is its purpose at the frontend? Can we derive it internally just like load_image? Having it is quite counter-intuitive in my point of view.

confluence: If a user wants to write a LEL expression for several files in the same base directory, it should be possible for them to specify the directory separately and use just relative file names in the expression (at least that's how I believe the frontend function uses the directory; this will have to be tested). But we have made the parameter optional -- so the user no longer *has to* fill this in (if all the paths in the expression are absolute and it doesn't matter what the directory is).

In the function for loading a hypercube, the directory parameter *never* matters (as far as we have tested), which is why we removed that parameter entirely.

kswang1029: in the frontend we can do 
`"dice_four.fits" + "/User/kswang/carta_image_pool/set_dice/dice_one.fits"` in a directory containing dice_four.fits (ie the first image)
but we cannot do 
`"/User/kswang/carta_image_pool/set_dice/dice_one.fits" + "dice_four.fits"` in the same folder. So based on this, I think we totally can derive the directory internally.

confluence: If you're asking why we can't derive it from a single path like in `open_image`, it's because we'd have to parse the entire LEL expression in the wrapper (we allow any expression, but we just treat it as a string which we pass to the frontend to be parsed).

kswang1029: how about in the docstring we explicitly say the images in the expression need to be all with absolution paths or all in the current directory? Then we don't need directory from users. Mixed usages like the examples I made are considered as edge cases which may or may not work. My intention here is to eliminate unclear user input.

kswang1029: what happens if we always set directory as pwd (does the frontend do so?)?

confluence: > in the frontend we can do `"dice_four.fits" + "/User/kswang/carta_image_pool/set_dice/dice_one.fits"` in a directory containing dice_four.fits (ie the first image) but we cannot do `"/User/kswang/carta_image_pool/set_dice/dice_one.fits" + "dice_four.fits"` in the same folder. So based on this, I think we totally can derive the directory internally.

Hmm. Do we know *why* that doesn't work? Is it because the browser's directory changes to the location of the first file, and then the directory is wrong for resolving the second file? This seems like undesirable behaviour. This means that if there's a mix of absolute and relative paths the LEL expression may not work no matter what directory is specified, depending on the order.

We can always set the directory to the "current" directory. But why should we remove the user's ability to use `open_LEL_image("A.fits + B.fits + C.fits", "path/to/those/files/")`, which should work (since all files are in the same directory)?

I'm also not sure whether we can use the session's `pwd` (as I thought we could). In the functions which have a separate path per file, we ensure that we can bypass the GUI browser's changing starting directory by converting relative paths to absolute paths (relative to our `pwd`, which doesn't change unless the user changes it). But if the frontend function changes the directory while processing the LEL expression, and calculates relative file paths using that directory, that's going to be a problem. I'm going to need to look more closely at how LEL expressions are handled.

It may be possible for us to parse the expression and rewrite the paths -- if we did that, we could ensure that the paths are all absolute and the expression always works (regardless of the starting directory and the order). But I would have to see how complex the syntax is.

kswang1029: Or, we rename path as abs_path, just to ask users to use absolute paths. In the scripting interface this is fine I think.

kswang1029: I think in the scripting interface "pwd" is not trivial to users. They need to explicitly use pwd to see the path. On the GUI, it is trivial in the file browser. Guessing this is the root of the problem/question.

I-Chenn: Since it's the only way for the circumstance that **_user_** only provide **_file name_** in `expression`. `load_image` also need user to provide `path` (the combination of `directory` and `file_name`) to derive.

I-Chenn:
@confluence: I have just modified the `new` function in `image.py` back to the status that not exposing it in be12659 as we discussed earlier.

carta/session.py
I-Chenn: Sorry, since the cleaning staff went to our office this afternoon for cleaning and I was out for my late lunch thus I just missed the whole discussion.

I think we still need that `directory` argument since:
> in the frontend we can do `"dice_four.fits" + "/User/kswang/carta_image_pool/set_dice/dice_one.fits"` in a directory containing dice_four.fits (ie the first image) but we cannot do `"/User/kswang/carta_image_pool/set_dice/dice_one.fits" + "dice_four.fits"` in the same folder. So based on this, I think we totally can derive the directory internally.

When you said _"in a directory containing dice_four.fits"_, you have actually **specified the `directory` argument with your file browser**, and hence the action of specifying `directory` argument is the exact standard action taken by the front-end UI usage.

Therefore, I think the current logic for `open_LEL_image` is reasonable: let user to decide whether to provide directory in `expression` argument or with `directory` argument (but mention not to mix two ways together in the description). If user wants to provide directory in `expression` argument, we clarify explicitly in the description that the input directory should be absolute directory.

confluence: I still want to investigate why the order matters (in the GUI browser) and if there's a way to fix this (in the wrapper or in both the GUI and the wrapper). It would be nice to fix this -- intuitively I would expect the directory to be used as the base directory for all file paths in the expression, and for the order not to matter.

acdo2002 (approved):
I have test three functions: open_LEL_image, open_image, and open_complex_image.
For the open_LEL_image, I tested with directory or absolute path, they are all working fine.
The testing script is in [here](https://github.com/acdo2002/carta-python-testing-examples/blob/dev/src/test/PR108_OpenImageLELComplex.py)

carta/image.py
```diff
@@ -42,39 +42,32 @@ def __init__(self, session, image_id, file_name):
...
-            Arithmetic expression to use if opening a complex-valued image.
+        image_arihmetic : boolean
+            Whether the file name should be interpreted as a LEL expression.
 
         Returns
         -------
         :obj:`carta.image.Image`
             A new image object.
         """
-        path = session.resolve_file_path(path)
```
confluence, line 74: I think I previously told you to move this out into the session functions, but I have realised that it makes more sense to keep it in here (we just need to call it on the `directory` parameter rather than the original `path` before it is split -- the effect should be the same). What this function actually _does_ is about to change (once the code to remove the custom pwd is merged in), but we still need it for relative paths.

carta/session.py
```diff
@@ -358,39 +359,58 @@ def set_starting_directory(self, path):
...
```
confluence: Once my changes are merged in, the default for `directory` here should be changed to `""` or `"."`, and the explicit handling of `None` in the body of the function should be removed. (This is in addition to moving the `resolve_file_path` step back into `Image.new`.)

confluence (commented):
Some minor changes to be made after the code for removing the custom pwd is merged in.

I-Chenn:
@confluence:
It takes me a while to resolve the conflicts due to the merge of #127. I haven't added the unit test of the new added function of this PR, but would you please take a look at my modification first and see if there's anything to be corrected before I add the test tomorrow morning?

confluence:
Your changes so far look fine to me -- but please merge dev again and resolve the conflicts in `constants.py` (you need to rename one class while preserving the new inheritance from `StrEnum`). This merge should add CI to this PR, and you will then have some automatic feedback. You still need to update the existing tests (to change various function signatures). Before you do that, please also move the `resolve_file_path` step back into `Image.new`, or the current image test will fail (since it assumes that `new` is performing this step).

I-Chenn:
There's still some problem at the `test_open_image` in 66533ed. Will try to resolve the problem tomorrow.

carta/image.py
```diff
@@ -59,8 +59,10 @@ def new(cls, session, path, hdu, append, complex, expression, make_active=True,
             Whether the image should be appended.
         complex : boolean
             Whether the image is complex.
-        expression : a member of :obj:`carta.constants.ArithmeticExpression`
-            Arithmetic expression to use if opening a complex-valued image. Ignored if the image is not complex.
+        image_arihmetic : boolean
```
confluence: There's a typo in the variable name here.

carta/image.py
```diff
@@ -73,6 +75,14 @@ def new(cls, session, directory, file_name, hdu, append, image_arithmetic, make_
         """
         command = "appendFile" if append else "openFile"
 
+        path = session.resolve_file_path(path)
+        directory, file_name = posixpath.split(path)
+
+        if complex and image_arithmetic:
+            file_name = f'{expression}("{file_name}")'
```
confluence: You unnecessarily moved this part of the expression logic into `new`. This part should *not* be here; it should only be in the session function for opening a complex image. This is only applicable to complex images. If you are opening a LEL expression, the expression parameter here is something completely different (an arbitrary string expression, not a constant which is used to construct the expression).

What you were previously passing to this function (a directory and a file name, where the file name could also be a string expression) was correct. Please revert most of these changes. The *only* thing that you should be moving in here is the `resolve_file_path` step, which you should be calling *on the directory parameter* that you are passing in here.

`open_image` and `open_complex image` should take a single path parameter and split it into a directory and filename, and `open_LEL_image` should take a separate directory and expression parameter. This was correct before.

carta/session.py
```diff
@@ -369,33 +367,33 @@ def open_complex_image(self, path, expression=ComplexExpression.AMPLITUDE, appen
...
-        path = self.resolve_file_path(path)
-        directory, file_name = posixpath.split(path)
-        expression = f'{expression}("{file_name}")'
-        return Image.new(self, directory, expression, "", append, True, update_directory=update_directory)
+        return Image.new(self, path, hdu, append, True, True, expression, update_directory=update_directory)
 
-    @validate(String(), String(), Boolean(), Boolean(), Boolean())
-    def open_LEL_image(self, expression, directory=".", append=False, update_directory=False):
+    @validate(String(), String(), String(), Boolean(), Boolean())
+    def open_LEL_image(self, expression, path=".", hdu="", append=False, update_directory=False):
```
confluence: As discussed above, the way you have changed these parameters and the way you are calling the new function from this function does not make sense and will not produce the correct result.

confluence (changes requested):

I-Chenn:
@confluence, @acdo2002:
I have:
1. Reverted ce428c9 and 66533ed.
2. Moved `resolve_file_path` to `Image.new`
3. Added unit test for `test_open_complex_image` and `test_open_LEL_image`.

The PR is available for review. Please review this again and see what can be improved.
Thank you.

carta/image.py
```diff
@@ -42,25 +40,25 @@ def __init__(self, session, image_id, file_name):
...
         hdu : string
             The HDU to open.
         append : boolean
             Whether the image should be appended.
-        complex : boolean
-            Whether the image is complex.
-        expression : a member of :obj:`carta.constants.ArithmeticExpression`
-            Arithmetic expression to use if opening a complex-valued image. Ignored if the image is not complex.
+        image_arithmetic : boolean
+            Whether the file name should be interpreted as a LEL expression. Ignored if the image is not complex.
```
confluence: Please remove `Ignored if the image is not complex.` -- this is no longer relevant.

carta/image.py
```diff
@@ -42,25 +40,25 @@ def __init__(self, session, image_id, file_name):
...
+        This method should not be used directly. It is wrapped by :obj:`carta.session.Session.open_image`, :obj:`carta.session.Session.open_complex_image` and :obj:`carta.session.Session.open_LEL_image`.
 
         Parameters
         ----------
         session : :obj:`carta.session.Session`
             The session object.
-        path : string
-            The path to the image file, either relative to the session's current directory or an absolute path relative to the CARTA backend's root directory.
+        directory : string
+            The directory containing the image file, or the base directory for the LEL arithmetic expression, as an absolute path relative to the CARTA backend's root directory.
```
confluence: This needs to be updated to the language used after the pwd change: `either relative to the session's current directory or an absolute path relative to the CARTA backend's root directory`

carta/session.py
```diff
@@ -340,45 +340,59 @@ def cd(self, path):
 
     # IMAGES
 
-    @validate(String(), String(r"\d*"), Boolean(), Constant(ArithmeticExpression), Boolean())
-    def open_image(self, path, hdu="", complex=False, expression=ArithmeticExpression.AMPLITUDE, update_directory=False):
-        """Open a new image, replacing any existing images.
+    @validate(String(), String(r"\d*"), Boolean(), Boolean())
+    def open_image(self, path, hdu="", append=False, update_directory=False):
```
confluence: The `make_active` flag is missing. It needs to be added to all of these functions (defaulting to `True`), because all of these functions can now append.

carta/session.py
```diff
@@ -340,45 +340,59 @@ def cd(self, path):
...
         """
-        return Image.new(self, path, hdu, False, complex, expression, update_directory=update_directory)
+        directory, file_name = posixpath.split(path)
+        return Image.new(self, directory, file_name, hdu, append, False, update_directory=update_directory)
 
-    @validate(String(), String(r"\d*"), Boolean(), Constant(ArithmeticExpression), Boolean(), Boolean())
-    def append_image(self, path, hdu="", complex=False, expression=ArithmeticExpression.AMPLITUDE, make_active=True, update_directory=False):
-        """Append a new image, keeping any existing images.
+    @validate(String(), Constant(ComplexExpression), Boolean(), Boolean(), Boolean())
+    def open_complex_image(self, path, expression=ComplexExpression.AMPLITUDE, append=False, update_directory=False):
```
confluence: This function needs the `make_active` flag.

carta/session.py
```diff
@@ -340,45 +340,59 @@ def cd(self, path):
...
+            Whether the image should be appended to existing images. By default this is ``False`` and any existing open images are closed.
+        update_directory : {3}
+            Whether the starting directory of the frontend file browser should be updated to the parent directory of the image. The default is ``False``.
+        """
+        directory, file_name = posixpath.split(path)
+        expression = f'{expression}("{file_name}")'
+        return Image.new(self, directory, expression, "", append, True, update_directory=update_directory)
+
+    @validate(String(), String(), Boolean(), Boolean(), Boolean())
+    def open_LEL_image(self, expression, directory=".", append=False, update_directory=False):
```
confluence: This function needs the `make_active` flag.

tests/test_image.py
```diff
@@ -101,22 +101,28 @@ def test_image_properties_have_docstrings(member):
...
-    (["subdir/image.fits", "", False, False, None], {},
+    (["subdir", "image.fits", "", False, False], {},
      ["openFile", "/my_data/subdir", "image.fits", "", False, False]),
     # Open a complex image
-    (["subdir/image.fits", "", False, True, AE.AMPLITUDE], {},
+    (["subdir", 'AMPLITUDE("image.fits")', "", False, True], {},
      ["openFile", "/my_data/subdir", 'AMPLITUDE("image.fits")', "", True, False]),
+    # Open a LEL image
+    (["subdir", "2*image.fits", "", False, True], {},
+     ["openFile", "/my_data/subdir", "2*image.fits", "", True, False]),
```
confluence: As far as this function is concerned, there is no difference between these two cases -- they're both just expressions. Only one of these test cases is needed.

tests/test_image.py
```diff
@@ -101,22 +101,28 @@ def test_image_properties_have_docstrings(member):
...
-    (["subdir/image.fits", "", True, False, None], {},
+    (["subdir", "image.fits", "", True, False], {},
      ["appendFile", "/my_data/subdir", "image.fits", "", False, True, False]),
     # Append a complex image
-    (["subdir/image.fits", "", True, True, AE.AMPLITUDE], {},
+    (["subdir", 'AMPLITUDE("image.fits")', "", True, True], {},
      ["appendFile", "/my_data/subdir", 'AMPLITUDE("image.fits")', "", True, True, False]),
+    # Append a LEL image
+    (["subdir", "2*image.fits", "", True, True], {},
+     ["appendFile", "/my_data/subdir", "2*image.fits", "", True, True, False]),
```
confluence: As above.

tests/test_session.py
```diff
@@ -115,16 +115,16 @@ def test_cd(session, mock_method, mock_call_action):
...
-    # Open complex image
-    (["subdir/image.fits"], {"complex": True},
-     ["subdir/image.fits", "", False, True, AE.AMPLITUDE], {"update_directory": False}),
+     ["subdir", "image.fits", "3", False, False], {"update_directory": False}),
     # Open plain image; update file browser directory
     (["subdir/image.fits"], {"update_directory": True},
-     ["subdir/image.fits", "", False, False, AE.AMPLITUDE], {"update_directory": True}),
+     ["subdir", "image.fits", "", False, False], {"update_directory": True}),
 ])
 def test_open_image(mocker, session, args, kwargs, expected_args, expected_kwargs):
```
confluence, line 132: All of these function tests need a test case for the `make_active` flag (which is currently missing).

tests/test_session.py
```diff
@@ -135,27 +135,39 @@ def test_open_image(mocker, session, args, kwargs, expected_args, expected_kwarg
...
+    # Open complex image (AMPLITUDE)
+    (["subdir/image.fits"], {"expression": CE.AMPLITUDE},
+     ["subdir", 'AMPLITUDE("image.fits")', "", False, True], {"update_directory": False}),
+    # Append complex image (PHASE)
+    (["subdir/image.fits"], {"expression": CE.PHASE, "append": True},
+     ["subdir", 'PHASE("image.fits")', "", True, True], {"update_directory": False}),
+    # Open complex image (REAL); update file browser directory
+    (["subdir/image.fits"], {"expression": CE.REAL, "update_directory": True},
+     ["subdir", 'REAL("image.fits")', "", False, True], {"update_directory": True}),
 ])
```
confluence, line 156: This needs a test case with no explicit expression parameter (to test the default).

confluence (changes requested):

carta/image.py
I-Chenn: OK, but I have modified this line as `Whether the file name should be interpreted as a complex expression or a LEL expression` since the parameter is `True` for both cases.
I thus also modified the `file_name` description to: `The name of the image file, complex expression or a LEL arithmetic expression`.

confluence: A complex expression is just a specific kind of LEL arithmetic expression. We have an additional session function to handle this special case, but the `new` function doesn't distinguish between them. `open_complex_image` constructs a LEL expression string from the complex expression constant and the file name.  There is no need to mention complex images here specifically.

I would actually like to rename the `ComplexExpression` enum, because it's not an expression (and the parameter with this enum value should not have the same name as the parameters with the string expressions). @kswang1029 would it make sense to call this a "complex part" or a "complex coordinate"?

I-Chenn: OK, I have remove complex image in the description in b5bffc5.

carta/session.py
```diff
@@ -360,16 +360,16 @@ def open_image(self, path, hdu="", append=False, make_active=True, update_direct
         directory, file_name = posixpath.split(path)
         return Image.new(self, directory, file_name, hdu, append, False, make_active=make_active, update_directory=update_directory)
 
-    @validate(String(), Constant(ComplexExpression), Boolean(), Boolean(), Boolean())
-    def open_complex_image(self, path, expression=ComplexExpression.AMPLITUDE, append=False, make_active=True, update_directory=False):
+    @validate(String(), Constant(ComplexComponent), Boolean(), Boolean(), Boolean())
+    def open_complex_image(self, path, expression=ComplexComponent.AMPLITUDE, append=False, make_active=True, update_directory=False):
```
confluence: Please also rename this `expression` parameter to `component`.

carta/session.py
```diff
@@ -360,16 +360,16 @@ def open_image(self, path, hdu="", append=False, make_active=True, update_direct
...
         """Open or append a new complex-valued image.
 
         Parameters
         ----------
         path : {0}
             The path to the complex-valued image file, either relative to the session's current directory or an absolute path relative to the CARTA backend's root directory.
         expression : {1}
-            The complex expression to use when opening the image. The default is :obj:`carta.constants.ComplexExpression.AMPLITUDE`.
+            The complex component to use when opening the image. The default is :obj:`carta.constants.ComplexComponent.AMPLITUDE`.
```
confluence: Please change `use` to `select` here.

confluence (changes requested):

carta/session.py
```diff
@@ -340,45 +340,65 @@ def cd(self, path):
...
+        append : {2}
+            Whether the image should be appended to existing images. By default this is ``False`` and any existing open images are closed.
+        make_active : {3}
+            Whether the image should be made active in the frontend. This only applies if an image is being appended. The default is ``True``.
+        update_directory : {4}
+            Whether the starting directory of the frontend file browser should be updated to the parent directory of the image. The default is ``False``.
+        """
+        directory, file_name = posixpath.split(path)
+        component = f'{component}("{file_name}")'
+        return Image.new(self, directory, component, "", append, True, make_active=make_active, update_directory=update_directory)
```
confluence: Please change the name of the parameter that you are passing to `new` back to `expression`. The parameter passed into this function is an enum value which represents the component. In line 381 it is used to create an expression, and that is passed into the `new` function. The purpose of the name change was to highlight the difference between these two things.

tests/test_session.py
```diff
@@ -135,27 +135,42 @@ def test_open_image(mocker, session, args, kwargs, expected_args, expected_kwarg
...
+    (["subdir/image.fits"], {"expression": CC.PHASE},
+     ["subdir", 'PHASE("image.fits")', "", False, True], {"make_active": True, "update_directory": False}),
+    # Append complex image (REAL)
+    (["subdir/image.fits"], {"expression": CC.REAL, "append": True},
+     ["subdir", 'REAL("image.fits")', "", True, True], {"make_active": True, "update_directory": False}),
+    # Open complex image (IMAG); update file browser directory
+    (["subdir/image.fits"], {"expression": CC.IMAG, "update_directory": True},
+     ["subdir", 'IMAG("image.fits")', "", False, True], {"make_active": True, "update_directory": True}),
+])
+def test_open_complex_image(mocker, session, args, kwargs, expected_args, expected_kwargs):
```
confluence, line 157: This test is now failing -- you need to update the name of the keyword parameter in the test cases.

confluence (changes requested):

confluence (approved):
This is almost ready to merge. 👍 There are a few minor punctuation issues in the docstrings, and I would like to revise some of the test cases -- I will make these changes before merging.

confluence:
@kswang1029 I think that this is ready for final testing.

kswang1029:
@confluence if I do
`img0 = session.open_LEL_image("AMPLITUDE('complex.image')", directory="/Users/kswang/set_QA_e2e_v22/",update_directory=False)`

where `"/Users/kswang/set_QA_e2e_v22/"` does not exist, the backend will crash. Do you think it is a backend issue? On GUI I think we do not have that freedom to control which directory to look for images (pwd it is) but now with python this becomes possible.

If we do
`img0 = session.open_LEL_image("AMPLITUDE('/Users/kswang/set_QA_e2e_v22/complex.image')", update_directory=False)` instead (again set_QA_e2e_v22 does not exist)
the backend survives with a popup error. 

Just in case this is useful, the macOS crash report is

> Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
> 0   carta_backend                 	       0x1004674fc carta::Session::OnOpenFile(CARTA::OpenFile const&, unsigned int, bool) + 204 (Session.cc:482)
> 1   carta_backend                 	       0x1004882ec carta::SessionManager::OnMessage(uWS::WebSocket<false, true, carta::PerSocketData>*, std::__1::basic_string_view<char, std::__1::char_traits<char>>, uWS::OpCode) + 3860 (SessionManager.cc:263)
> 2   carta_backend                 	       0x10048f318 ofats::any_detail::any_invocable_impl<void, false, uWS::WebSocket<false, true, carta::PerSocketData>*, std::__1::basic_string_view<char, std::__1::char_traits<char>>, uWS::OpCode>::call(uWS::WebSocket<false, true, carta::PerSocketData>*, std::__1::basic_string_view<char, std::__1::char_traits<char>>, uWS::OpCode) + 24 (MoveOnlyFunction.h:247) [inlined]
> 3   carta_backend                 	       0x10048f318 ofats::any_invocable<void (uWS::WebSocket<false, true, carta::PerSocketData>*, std::__1::basic_string_view<char, std::__1::char_traits<char>>, uWS::OpCode)>::operator()(uWS::WebSocket<false, true, carta::PerSocketData>*, std::__1::basic_string_view<char, std::__1::char_traits<char>>, uWS::OpCode) + 24 (MoveOnlyFunction.h:354) [inlined]
> 4   carta_backend                 	       0x10048f318 uWS::WebSocketContext<false, true, carta::PerSocketData>::handleFragment(char*, unsigned long, unsigned int, int, bool, uWS::WebSocketState<true>*, void*) + 628 (WebSocketContext.h:93)
> 5   carta_backend                 	       0x10048e35c bool uWS::WebSocketProtocol<true, uWS::WebSocketContext<false, true, carta::PerSocketData>>::consumeMessage<6u, unsigned char>(unsigned char, char*&, unsigned int&, uWS::WebSocketState<true>*, void*) + 928 (WebSocketProtocol.h:328)
> 6   carta_backend                 	       0x10048de74 uWS::WebSocketProtocol<true, uWS::WebSocketContext<false, true, carta::PerSocketData>>::consume(char*, unsigned int, uWS::WebSocketState<true>*, void*) + 148 (WebSocketProtocol.h:424)
> 7   carta_backend                 	       0x10048dd7c auto uWS::WebSocketContext<false, true, carta::PerSocketData>::init()::'lambda'(auto*, char*, int)::operator()<us_socket_t>(auto*, char*, int) const + 144 (WebSocketContext.h:286)
> 8   carta_backend                 	       0x100536c10 us_loop_run + 204 (epoll_kqueue.c:147)
> 9   carta_backend                 	       0x10048ad6c uWS::Loop::run() + 4 (Loop.h:159) [inlined]
> 10  carta_backend                 	       0x10048ad6c uWS::run() + 12 (Loop.h:176) [inlined]
> 11  carta_backend                 	       0x10048ad6c uWS::TemplatedApp<false>::run() + 12 (App.h:393) [inlined]
> 12  carta_backend                 	       0x10048ad6c carta::SessionManager::RunApp() + 260 (SessionManager.cc:593)
> 13  carta_backend                 	       0x1003db35c main + 3468 (Main.cc:189)
> 14  dyld                          	       0x1a4e1ff28 start + 2236

kswang1029 (approved):

PR #113, opened 2023-05-15
Icchen/81 high resolution export

### LINKED ISSUE #81: support a resolution scaling factor for export-as-png
The frontend supports three resolution scaling options as 100% (as is on screen), 200%, and 400%, if users need a higher resolution PNG image for presentation purposes. Session.save_rendered_view() function could be enhanced to support this. 

Currently we can use Session.set_view_area() to a desired PNG size. Having the scaling factor would be a handy if users are in the interactive mode. For headless mode, Session.set_view_area() alone should be sufficient.

I-Chenn:
@confluence 
This draft PR is to resolve #81.

An argument `image_ratio` and several steps are added into `save_rendered_view` to output png-image with different image ratio:
Setting `setIsExportingImage` to `True` enables setting output image ratio by calling `setImageRatio`. After writing the data to the file, we set `setIsExportingImage` to `False` and reset the image ratio to `1` with `resetImageRatio`.

The new-configured function does work for a new opened session. With the script like:
```
session.save_rendered_view(file_name="/home/icchen/Pictures/test/test_4.png", image_ratio=4)
```
I am able to generate a reolution = 400% png image like:
test_4
Nevertheless, there's something really odd: after a period of time (say 20 - 30 mins) without touching anything, the exact same execution will then generate something like:
test_4a
The image is still with the same 400% resolution in size, but the output content seems to lock to 100% resolution.

I still have no idea what's the reason that causes the phenomena, but I will discuss with @YuHsuan-Hwang and see if there's any mechanism at frontend will causes that.

### DIFF
--- a/carta/session.py
+++ b/carta/session.py
@@ -799,8 +799,8 @@ class Session:
 
     # SAVE IMAGE
 
-    @validate(NoneOr(Color()))
-    def rendered_view_url(self, background_color=None):
+    @validate(NoneOr(Color()), Number())
+    def rendered_view_url(self, background_color=None, image_ratio=1):
         """Get a data URL of the rendered active image.
 
         Parameters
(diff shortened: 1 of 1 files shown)


confluence (changes requested):
The image ratio parameter should be added to all three functions for retrieving data -- `rendered_view_url`, `rendered_view_data` as well as `save_rendered_view`. The user could access any of these stages directly. `setImageRatio`, `resetImageRatio` and `setIsExportingImage` should be called from the innermost function (`rendered_view_url`).

I-Chenn:
@confluence I made some modifications according to your suggestions, the issue mentioned above (lock to 100% resolution issue after 20 - 30 mins) is still appearing. I firstly thought it probably due to: we reset the `image_ratio` to 1 not until the data is written into the png file (and that's the reason I want to call `resetImageRatio` in `save_rendered_view` after the writing file line). Nevertheless, it seems that the issue still appears if I do so. Will keep investigating for the cause of the issue.

I-Chenn:
@confluence, another issue is that we need to restrict the `image_ratio`. Do you think if it's proper to set 5 as the upper limit for `image_ratio`?

Issue #109, opened 2023-05-10
Add a framework for high-level function unit tests


confluence:
~We need a dummy protocol class that outputs function calls in a format which can be checked for correctness. The test framework should provide dummy session and image objects which use this protocol class.~ We actually don't need to write any of this ourselves, because `pytest` has a nice mocking plugin.

confluence:
Implemented in #116.

Issue #57, opened 2022-09-14
allowing angular size as input in Image.set_zoom()


kswang1029:
currently `Image.set_zoom()` takes the zoom factor as the input. This is not intuitive in general use cases and it depends on the screen resolution. The better inputs should be 1) in pixels to fill up the image view and 2) in angular size to fill up the image view. Examples:

`Image.set_zoom('123pix')`
`Image.set_zoom('20arcsec')`
`Image.set_zoom('4arcmin')`
`Image.set_zoom('2deg')`

This may be straightforward once https://github.com/CARTAvis/carta-frontend/issues/1176 is implemented (otherwise we will need to handle image<->world coordinates transformation...)

We can still keep the zoom factor as one kind of inputs.

kswang1029:
this is now possible with the frontend based on https://github.com/CARTAvis/carta-frontend/pull/2081

Jordatious:
Would this just Zoom, or change the actual size / shape of the viewer window? I think both are needed, meaning it should support two inputs for reshaping. An interactive (i.e. not headless) scripting session would support manual resizing of the viewer window.

Jordatious:
At this point, the user could address this with trial and error, assuming any real number is accepted as input.

confluence:
Implemented in #116.

Issue #56, opened 2022-09-14
allowing world coordinates as input in Image.set_center()


kswang1029:
Currently, `Image.set_center()` takes two numbers as the image coordinates as the input. For a better usability, we will need to also support inputs as world coordinates such as:

`Image.set_center('12h34m.56.789s', '-10d20m30.45s')`
`Image.set_center('123.3456deg', '-12.3456deg')`

This may be straightforward once https://github.com/CARTAvis/carta-frontend/issues/1176 is implemented (otherwise we will need to handle image<->world coordinates transformation...)

kswang1029:
this is now possible with the frontend based on https://github.com/CARTAvis/carta-frontend/pull/2081

Jordatious:
Is the current input pixel units then?

kswang1029:
> Is the current input pixel units then?

Yes

Jordatious:
Ok thanks. The user could get around this with other WCS libraries, but certainly this would be a nice additional feature for CARTA.

confluence:
Implemented in #116.

PR #116, opened 2023-05-17
Add use of WCS units to center and zoom functions

confluence:
This PR implements #56, #57, and #109.

* `Image.set_center` now accepts either world coordinates or image coordinates.
  * The validation permits a wider range of formats than the GUI pan and zoom dialog, including those requested in the example snippet in #56. These formats are normalized to the forms accepted by the frontend. Numbers or numeric strings without units are assumed to be in degrees, for consistency with the GUI.
  * A limitation is that the parameter formats must match the currently set overlay number formats (this is required by the frontend). We can detect a mismatch, but we can't automatically correct the settings because 1) we can't tell if a HMS string is supposed to be HMS or DMS, 2) we can't deduce a coordinate system from a format, and 3) we don't know if the user intends to change the coordinate system or set a custom number format. So it has to be the user's responsibility to ensure this.
  * `set_center` does have an optional parameter to set the system (for consistency with the dropdown in the GUI dialog), and I have added high-level functions to `Session` to make setting or clearing a custom number format more user-friendly.
  * Another limitation is that image and world coordinates can't be combined. We detect this and fail explicitly.
  * We also fail if world coordinates are provided and the image has no WCS info.
* The old `Image.set_zoom` has been renamed to `Image.set_zoom_level`.
* `Image.zoom_to_size_x` and `Image.zoom_to_size_y` are used to fit the specified dimension to the size provided, and accept both pixel sizes and angular sizes.
  *  We accept a wider range of unit names than the frontend, and normalize them (to `"` for arcsec, `'` for arcmin, and `deg` for degrees). A number or numeric string with no units is assumed to be in arcsec (for consistency with the frontend).
  * We fail if an angular size is provided and the image has no WCS info.
* This PR also contains unit tests for the high-level functionality added in this PR, as well as a basic framework for writing unit tests with `pytest`. Once this is merged, tests for existing high-level functions can be filled in.
  * The unit test directory has been added to the format scripts.
* `CoordinateSystem` enum values have been corrected to match the frontend.
* The `String` validation class now has an integer parameter for arbitrary regex flags instead of a boolean parameter for ignoring the case. Existing code was using the parameter like this already (it worked by accident!).
* Macro objects can now be compared (needed for unit tests).
* Utility and validation classes for parsing and validating sizes and coordinates have been added and can be reused in other functions.

I made some decisions about the allowed / disallowed input formats which are up for discussion -- these can be changed relatively easily. For example:
* the full list of allowed units and formats
* what numbers without units default to in each function
* spaces not being allowed before `'` and `"`, but being allowed before multi-character unit words
* h/d, m and s are all optional in hms / dms format. All three colons must be present if the colon separator format is used (this is consistent with the frontend). If letters are used, both number and letter must be included or excluded together. This means that `""` is a valid coordinate which is evaluated as `"::"`.
* More generally, you can see what inputs are allowed by looking at the validation tests, and how they are transformed before being sent to the frontend by looking at the util tests.

To be completed:
- [x] Sample validation class tests
- [x] Session function for retrieving number format
- [x] Tests for this function and the function for retrieving the coordinate system
- [x] Basic e2e sanity check
- [x] Check the rendering of the documentation


I-Chenn (approved):
The new added functions work without any problem in my attempt of test.

confluence:
I wasn't entirely happy with the way that the parsing of the different formats was arranged internally, so I did a bit of refactoring to clean up the design and make the separation and code reuse more logical. The user-facing API is unchanged. While updating the tests I found some with duplicate names which I think never ran, and fixed them.

carta/session.py
```diff
@@ -495,6 +495,16 @@ def set_coordinate_system(self, system=CoordinateSystem.AUTO):
         """
         self.call_action("overlayStore.global.setSystem", system)
 
+    def coordinate_system(self):
```
kswang1029, line 498: shall we make it clear between native coordinate system (as defined in header) and "rendered" coordinate system?

carta/util.py
```diff
@@ -129,3 +134,214 @@ def split_action_path(path):
...
+        m = re.match(cls.UNIT_REGEX, value, re.IGNORECASE)
+        if m is None:
+            raise ValueError(f"{repr(value)} is not in a recognized pixel format.")
+        return m.group(1)
+
+
+class AngularSizeString:
+    """Parses angular sizes."""
+
+    NORMALIZED_UNIT = {
```
kswang1029: I suggest we also support 
1. milliarcsec (10^-3 arcsec for VLA/ALMA etc)
2. microarcsec (10^-6 arcsec for VLBI, EHT etc)

carta/util.py
```diff
@@ -129,3 +134,214 @@ def split_action_path(path):
...
+        "arcmin": "'",
```
kswang1029: we can consider to also support arcsecond and arcminute. Sometimes, we also use "asec" for arcsec and "amin" for arcmin.

carta/util.py
```diff
@@ -129,3 +134,214 @@ def split_action_path(path):
...
+        "DEGREE_UNIT": rf"^-?(\d+(?:\.\d+)?)\s*({'|'.join(DEGREE_UNITS)})$",
+        "HMS_COLON": r"^-?\d{0,2}:\d{0,2}:(\d{1,2}(\.\d+)?)?$",
+        "HMS_LETTER": r"^(?:(-?\d{1,2})h)?(?:(\d{1,2})m)?(?:(\d{1,2}(?:\.\d+)?)s)?$",
+        "DMS_COLON": r"^-?\d*:\d{0,2}:(\d{1,2}(\.\d+)?)?$",
+        "DMS_LETTER": r"^(?:(-?\d+)d)?(?:(\d{1,2})m)?(?:(\d{1,2}(?:\.\d+)?)s)?$",
+        "DECIMAL": r"^-?\d+(\.\d+)?$",  # No units = degrees
+    }
+
+    @classmethod
+    def valid(cls, value):
```
kswang1029, line 336: wonder if we need to validate value range? For example, for "hms" convention, `0 <= h < 24`, `0 <= m < 60`, and `0 <= s < 60`. For "dms", `-90 <= d <= 90`,`0 <= m < 60`, and `0 <= s < 60`. For "d" as "longitude", `0 <= d < 360` and as "latitude" `-90 <= d <= 90`

tests/test_image.py
```diff
@@ -0,0 +1,221 @@
...
+@pytest.mark.parametrize("y", [0, 10, 19])
+def test_set_center_valid_pixels(image, mock_property, mock_call_action, x, y):
+    mock_property("width", 20)
+    mock_property("height", 20)
+
+    image.set_center(f"{x}px", f"{y}px")
+    mock_call_action.assert_called_with("setCenter", float(x), float(y))
+
+
+@pytest.mark.parametrize("x,y,x_fmt,y_fmt,x_norm,y_norm", [
```
kswang1029, line 150: might be good to add test cases like (30d, -30d) where the latitude coordinate is a negative number. Need unit tests for this case too.

kswang1029 (changes requested):
@confluence I just had some initial code comments. Will proceed user testing.

carta/session.py
confluence: This is mostly unambiguous because it's a function on the session, not the image (and will probably soon move to a separate image viewer class). But we can explain this in more detail in the docstring.

carta/util.py
confluence: This is doable; I'll need to do a bit of refactoring to normalize these to one of the units accepted by the frontend. Are there any other aliases of these unit names that you would like to support?

carta/util.py
confluence: These are all very easy to add.

carta/util.py
confluence: Also doable. We would need to be aware of both the coordinate system and which coordinate is which, so this probably needs to be done inside the `set_center` function.

As far as I can tell, the frontend currently doesn't do any validation beyond the format -- it attempts to transform whatever values are provided and applies the result if it's valid.

tests/test_image.py
confluence: This would probably be in the unit test for `set_center`.

carta/util.py
confluence: Actually, for HMS and DMS we can do a more specific check in the util class. We'll just need the system and the axes for the degree ranges.

kswang1029:
@confluence here is one issue I spotted. If we have two images loaded (not matched) and they both have equatorial coordinate system defined as the native one, then if we try to set_center using GALACTIC coordinate to the non active image, the result will be incorrect. Example script is shown below (let me know if you need the two test images I used).

```
from carta.session import Session
from carta.constants import CoordinateSystem
import time

url_with_token = 'http://localhost:3002'
session_id = 3055419803

image_file_3 = '/Users/kswang/carta_image_pool/set_herschel_pacs_spire/spire500_ext.fits'
image_file_1 = '/Users/kswang/carta_image_pool/set_HD163296/HD163296_C18O_2-1.fits'


##############################################################

session = Session.interact(url_with_token, session_id, debug_no_auth=True)
img0 = session.open_image(image_file_3)
img1 = session.append_image(image_file_1)


print(img0.valid_wcs)
img0.set_center("7:09:00", "-11:12:37")
img0.set_zoom_level(2.0)

time.sleep(5)

img0.make_active() # try to comment out this to see the issue

img0.set_center(224.791, -1.733, system=CoordinateSystem.GALACTIC)
img0.set_zoom_level(4.0)


time.sleep(5)

session.save_rendered_view('test.png')
session.close()
```


https://github.com/CARTAvis/carta-python/assets/20819712/06664405-bd59-462d-87ae-d0c4fe7212dd

If we make image active first before set_center, it works. However, the axis labels are not rendered correctly.

https://github.com/CARTAvis/carta-python/assets/20819712/df3eba80-cddc-42cd-82d5-c84d2fceca8a

carta/util.py
confluence: I added these, together with "arcseconds" and "arcminutes" ("degrees" is also supported, but only these full words can be plural, not any of the abbreviations).

carta/util.py
confluence: I added these (together with the full words, and `mas` and `µas`). I also added `uas`, since I know that `u` has sometimes been used as an ASCII approximation for `µ`. It does turn up in search results, but possibly most of these results are just badly OCRed academic papers. If nobody actually uses this, I can take it out.

carta/util.py
confluence: @kswang1029:

1. Should the latitude range for degrees not be `-90 <= d <= 90`?
2. For HMS and DMS, I assume  that e.g. H can only be 24 if M and S are zero (i.e. 24:12:34.5 = not valid). Is that correct?
3. What if the user sets DMS as the custom number format for a longitude axis? I understand that this is not the convention, but nothing is preventing it. Should we then validate that `0 <= d < 360`, as for decimal degrees?
4. Am I correct in assuming that longitude = right ascension = x axis and latitude = declination = y axis in all coordinate systems *except for ecliptic* in which case latitude = x axis and right ascension = y axis? And does it still make sense for the user to provide the coordinates in (x, y) order in the latter case?

kswang1029: > @kswang1029:
> 
> 1. Should the latitude range for degrees not be `-90 <= d <= 90`?

When the coordinate system is galactic or ecliptic, we express the axis as Galactic/Ecliptic longitude and Galactic/Ecliptic latitude. Mostly, we use decimal degrees as the expression of longitude and latitude. For longitude, the range is 0 deg <= longitude < 360 deg, and for latitude, the range is -90 deg <= latitude <= 90 deg. 


> 2. For HMS and DMS, I assume  that e.g. H can only be 24 if M and S are zero (i.e. 24:12:34.5 = not valid). Is that correct?

Correct, the HMS (hour-minute-second) range is like 0h:0m:0s,  0h:0m:1s, ..., 23h59m59.999999....s. The DMS  (degree-arcmin-arcsec) range is -90d:0m:0s, ... -89d0m0s, ..., +90d0m0s. These are all sexagesimal. 

> 3. What if the user sets DMS as the custom number format for a longitude axis? I understand that this is not the convention, but nothing is preventing it. Should we then validate that `0 <= d < 360`, as for decimal degrees?

Usually for ICRS, FK5 and FK4 (ie equatorial system), we use the HMS-DMS convention (sexagesimal). But it is totally fine to convert sexagesimal format to decimal degree format. For GALACTIC and ECLIPTIC, we use decimal degrees convention. Rarely we use sexagesimal format for these two systems (but it is valid to do such conversion).

> 4. Am I correct in assuming that longitude = right ascension = x axis and latitude = declination = y axis in all coordinate systems _except for ecliptic_ in which case latitude = x axis and right ascension = y axis? And does it still make sense for the user to provide the coordinates in (x, y) order in the latter case?

AST has its logic on where to map the longitude/latitude or RA/DEC axes to the canvas xy axes if the field has extra rotation. Assuming there is no extra field rotation, for ICRS, FK5 and FK4 systems, North is up and East is left. Then x is RA and y is DEC. For GALACTIC and ECLIPTIC systems without extra field rotation, North is up and East is left too. So x is longitude and y is latitude. The tricky part is, if we select GALACTIC grid to render from a native FK5 system (or vice versa), we will see the grid is rotated (hence we may see unexpected longitude/latitude to x/y mapping due to AST).

confluence: Following up on our previous conversation, I also checked what happens in the GUI if you select _HMS_ as the format for _latitude_. This appears to be valid, and negative values are converted to the expected negative HMS values, which suggests that in this case the range should be -6:00:00 to 6:00:00.

kswang1029: Hmmm I would suggest we fix this behavior of the frontend. 🤔 showing HMS for latitude is not right.

confluence: That may be tricky. The preferences are for the X and Y axis, and as we discussed, which axis is which depends on how AST decides to display that image.

confluence: Fortunately I was overthinking this (I misread this part of your reply). I see now that the orientation of X and Y *also* changes, so X is always longitude and Y is always latitude. That should make it easier to implement the ranges (and also to disable specific custom number options in the frontend).

confluence: I also misread the HMS range -- I see that 24 is excluded from the hour range.

confluence:
I'm temporarily moving this back to draft so that I can push some incomplete changes while I'm refactoring. I will move it back out once it's ready for review again.

carta/image.py
```diff
@@ -426,23 +426,103 @@ def set_polarization(self, polarization, recursive=True):
...
+            self.session.set_coordinate_system(system)
+
+        x_is_pixel = PixelValue.valid(str(x))
+        y_is_pixel = PixelValue.valid(str(y))
+
+        if x_is_pixel and y_is_pixel:
+            # Image coordinates
+            x_value = PixelValue.as_float(str(x))
+            y_value = PixelValue.as_float(str(y))
+            if 0 <= x_value < self.width and 0 <= y_value < self.height:
```
kswang1029: This is minor but I just realized, in CARTA we refer the _center_ of the pixel at the bottom-left corner as (0, 0). This means, for example in x direction we have 100 pixels, the _valid_ range should be -0.5 <= x < 99.5 (or -0.5 <= x <= 99.5?).

carta/image.py
```diff
@@ -426,23 +426,103 @@ def set_polarization(self, polarization, recursive=True):
...
+        Parameters
+        ----------
+        size : {0}
+        """
+        size = str(size)
+        if PixelValue.valid(size):
+            self.call_action("zoomToSizeX", PixelValue.as_float(size))
+        else:
+            if not self.valid_wcs:
+                raise ValueError("Cannot parse angular size. This image does not contain valid WCS information.")
```
kswang1029: IIRC, if the image is not have a valid WCS, we render the xy axes as "linear". So maybe we can add a hint to use image coordinate (ie pixel) as the size input.

carta/image.py
```diff
@@ -426,23 +426,103 @@ def set_polarization(self, polarization, recursive=True):
...
+            self.call_action("zoomToSizeXWcs", str(AngularSize.from_string(size)))
+
+    @validate(Size())
+    def zoom_to_size_y(self, size):
```
kswang1029: Wonder if we can combine zoom_to_size_x and zoom_to_size_y as zoom_to_size and have an argument as "direction" (or "dimension" or "axis"?) for x or y? This would simplify the code repeatition here.

carta/image.py
confluence: I decided to make the range symmetrical -- if the user can center the bottom-left corner, they should also be able to center the other corners.

Does it make sense to apply a range limit at all? The user can specify WCS coordinates which are outside the image area, right? Should they not be able to do the same here?

carta/image.py
confluence: Good idea. I already added a `SpatialAxis` enum for the `set_center` range validation, so this is a simple change.

carta/image.py
kswang1029: Right, world coordinate can be outside the image. So guess we can remove the range check for consistency and potentially enable some use cases.

confluence:
As discussed, I will merge this as-is so that we can start using the testing framework, and create a new issue for the inactive image problem.

PR #127, opened 2023-07-12
Remove custom wrapper pwd

### LINKED ISSUE #126: Eliminate custom pwd
Once [frontend #2188](https://github.com/CARTAvis/carta-frontend/pull/2188) is merged, it will be possible to open an image from the wrapper without automatically changing the frontend's current directory to the image's parent directory. The custom wrapper-side pwd and all path transformations that use it (which were introduced as a workaround for this) should then be removed. The wrapper's `pwd` and `cd` should directly show and change the normal GUI browser's starting directory. It should be possible to pass unmodified relative paths to functions for opening images.

confluence:
Fixes #126. We no longer maintain a separate local pwd in the wrapper; we stay synchronised with the GUI file browser. This means that using `session.cd` updates the GUI browser's current dir, and vice versa. We still need the function for resolving file paths (for relative paths).

I have added a normalization step to e.g. turn "foo/bar/../../baz" into "baz", so that using e.g. `session.cd("..")` doesn't lead to increasingly convoluted paths, but this might interact weirdly with people's symbolically linked directories. I could take this out or make it optional, but I don't know how much of an issue it's likely to be.

I have also added unit tests for these functions.

### DIFF
--- a/carta/constants.py
+++ b/carta/constants.py
@@ -1,6 +1,6 @@
 """This module provides a collection of enums corresponding to various enumerated types and other literal lists of options defined in the frontend. The members of these enums should be used in place of literal strings and numbers to represent these values; for example: ``Colormap.VIRIDIS`` rather than ``"viridis"``. """
 
-from enum import Enum, IntEnum, auto
+from enum import Enum, IntEnum
 
 
 # TODO make sure the __str__ is right for all the string values
--- a/carta/image.py
+++ b/carta/image.py
@@ -42,7 +42,7 @@ class Image:
         self._frame = Macro("", self._base_path)
 
     @classmethod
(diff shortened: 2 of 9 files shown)


kswang1029 (approved):
This looks good. Tested the options `make_active` when appending images and `update_directory` when opening/appending images. Also tested the behavior of `cd`, `ls`, `pwd` in the session object. Code looks fine. 👍

I-Chenn (approved):
The added options `make_active` and `update_directory` works well when tested. The modified function `resolve_file_path`, `pwd` and `cd` are also without problem.
This commit may cause conflict with the not-yet-merged PR #108, will work on that branch to resolve the conflicts

PR #129, opened 2023-07-17
CI: branch name change

confluence:


confluence:
I had to change the branch name (the default branch variable probably evaluates to `main`).

There's a breaking change in the way that enums are stringified in 3.11. I added a fix and did a bit of refactoring in `constants`.

PR #136, opened 2023-08-16
Basic region support

### LINKED ISSUE #80: support regions/annotations
For the publication purpose, we need to support region/annotation file loading so users can use it to decorate their images for presentation (such as an arrow plus a text to point to a target source or a set of point markers to mark target sources, etc)

once https://github.com/CARTAvis/carta-frontend/pull/2019 and https://github.com/CARTAvis/carta-backend/pull/1229 are merged annotation region will be supported.

confluence:
This PR implements #80 and adds basic region support to the high-level wrapper. This PR depends on multiple frontend changes made in [#2231](https://github.com/CARTAvis/carta-frontend/pull/2231), so that should be merged first.

Required before this can be moved out of draft:
- [X] Fill in missing region and annotation functions (lock, focus, style, modify dimensions, etc.)
- [X] Add support for WCS coordinates and sizes (requires retroactive change to some existing code for consistency)
- [X] Fill in missing docstrings
- [X] Add basic tests
- [X] Factor out `BasePathMixin` into a separate PR (it should be merged first and used to refactor the existing code to use the same composition pattern as the region functionality)
- [X] Manually inspect the generated documentation for bugs.

Outstanding issues to be addressed before merging

- [X] Make sure that the custom code that patches missing frontend features (size, etc.) is consistent with the dev frontend updates.

This PR should be tested against the latest release version of CARTA. There will be a separate PR to remove the custom code in response to the dev frontend fixes, once this is merged and a main branch is created.


confluence:
This is ready for review. The best way to see the full range of available features is probably to build the docs locally and browse the package reference.

`img.regions` is a reference to the region set, and can be used to:
* list, get references to, and clear existing regions
* create regions of different types
* import and export regions

Methods on `img.regions` which create and retrieve regions return instances of `Region` subclasses. These objects can be used to get and set region properties, and delete individual regions. Some core methods are shared by all region types, and some are specific to particular types. For example, some regions have control points which are all vertices, some have only a start and end, and some have a center and a size. Some regions have the concept of a rotation, and some do not. The wrapper attempts to provide an interface which is as consistent as possible across different region types.

Outstanding frontend issue (which is mostly mitigated in the wrapper but will affect snippets): as described in [#2246](https://github.com/CARTAvis/carta-frontend/issues/2246), the way that the frontend handles the size of the ruler and other region types with endpoints is not consistent. This functionality is not exposed to the user in the GUI, but the discrepancies are apparent when snippets are used to e.g. set the size on regions directly. Fixing this in the frontend shouldn't affect the interface presented to the user, but would require internal changes to various calculations.

I have implemented additional functionality in the wrapper's `RulerAnnotation` class (following the current frontend implementation for line and vector regions) so that its behaviour matches the behaviour of those other types exactly. I have also overridden the `set_size` method for all the endpoint classes to negate the x and y values, because in the raw frontend code if you set the size to (a, b) and then retrieve the size you get (-a, -b) back.

However, some issues remain:
* I opted to negate the `set_size` input rather than the output of `size` because it was a smaller departure from the raw frontend representation, but I think that the opposite change makes more sense. If I draw a vector pointing in e.g. a positive X direction, I would expect the X coordinate of the size to be positive. Currently the opposite is true, and this seems counterintuitive.
* I have added rotation support to the ruler class (to match line and vector behaviour), but this rotation value only exists in the wrapper (in the GUI it will be displayed as zero).

Other observations:
* Numeric size and coordinate values are now unambiguously pixels, and string values are unambiguously WCS coordinates or angular sizes.
* A lot of the region functions take `(a, b)` tuples as parameters. This was a design decision to make it easier to validate point values.
* Many functions accept parameters in either WCS coordinates / angular sizes or pixels. These are converted to pixels before being sent to the frontend.
* There is a point class which is used internally to transform points between tuples and dictionaries. The user always enters tuples and receives tuples back (because this is the simplest and cleanest interface), but points sent to the frontend are converted to dictionaries with x and y keys, because this provides the correct JSON serialization for deserialization into Point2D in the frontend.
* The `set_size` method for compass annotations is overridden to use the frontend's `setLength`, because the frontend's `setSize` does not resize compass annotations. The wrapper's `set_size` sets the length to the smaller of the two dimensions provided. There is a separate `set_point_length` method for this type which can be called with a single size parameter.

I'm probably forgetting other things -- please let me know if anything else is confusing.

confluence:
As usual, the unit tests only test the wrapper's behaviour. I haven't added extensive tests for the extra ruler functionality, because it's taken straight from the frontend, and may be removed if the frontend code is extended to this region type. E2E tests for checking that the wrapper + frontend produces consistent results for the different region types would be useful.

confluence:
Note to self: the behaviour of line-like region `set_size` should be updated to be consistent with https://github.com/CARTAvis/carta-frontend/pull/2280.

(I.e. the wrapper behaviour with the last release version of CARTA should be consistent with the behaviour changes in that PR; it will be updated to work with the actual changes in the PR after this PR is merged and a main branch is created.)

confluence:
The functionality of the line-like regions (getting / setting the size /rotation / center of the line region, line annotation, vector annotation, and ruler annotation) should now be the same in this wrapper + the release version of CARTA as it is in the development version of the frontend (once [#2280](https://github.com/CARTAvis/carta-frontend/pull/2280) is merged, when tested with snippets). (A known difference is that for the ruler the rotation may sometimes be a float where the rotation for an identical line region is an integer -- but the values should be the same.)

confluence:
@kswang1029 I wonder if we should adopt [the `elements` nomenclature](https://github.com/CARTAvis/carta-frontend/pull/2335) here as well.

kswang1029:
> @kswang1029 I wonder if we should adopt [the `elements` nomenclature](https://github.com/CARTAvis/carta-frontend/pull/2335) here as well.

maybe not. Where in your mind should be?

confluence:
> maybe not. Where in your mind should be?

The wrapper is currently using "region" as a generic term for region or annotation, although in some places both are mentioned explicitly. I was wondering if we were moving away from that in the frontend, but I guess this change only affects one part of the UI (and e.g. we haven't renamed the region list). So what we have now is probably clear enough.

kswang1029:
> > maybe not. Where in your mind should be?
> 
> The wrapper is currently using "region" as a generic term for region or annotation, although in some places both are mentioned explicitly. I was wondering if we were moving away from that in the frontend, but I guess this change only affects one part of the UI (and e.g. we haven't renamed the region list). So what we have now is probably clear enough.

I see. Let’s keep things as they are now. Using “element” in the region file dialog is just to simply the info to one line. Otherwise we need two lines for region and annotation, respectively.

Jordatious:
Hi @confluence, I'd like to have a little play with this. Do you have any documented examples or setting WCS regions, or could you provide some? I can get a vague idea from the docs and docstrings, but it would be useful to get started with some examples!

confluence:
Hi, @Jordatious . I don't think there's currently any documentation for this outside the API docs. The general idea is:

* You access the objects for existing regions with the `list` and `get` methods on `your_image.regions`
* You create new regions with `add_point`, `add_rectangle`, and so on. The base `add_region` isn't intended for interactive use like this, but it's available in case you want to iterate over a whole bunch of region properties and add them programmatically. Unlike in the GUI, there's only one method for regions and annotations with the same shape (there's an `annotation` flag, which is false by default). These methods return the region object (the way that the session functions for opening images return an image object).
* You can interact with existing regions using their objects -- you should be able to inspect or change any property. This includes changing the control points (the different region types have different helper methods for this, because the control points are interpreted differently). The docstrings should specify which control point parameters accept WCS coordinates / sizes or pixels, and which have to be one or the other. IIRC there's currently no way to change the type of a region, but if there's demand for this I could look into it (in the meantime, you can manually create a new region using properties taken from an existing region, and then remove the existing region).
* You can import regions from a file with `your_image.regions.import_from`. You can export multiple regions with the corresponding `export_to`, and export a single region with the `export_to` on the region object.

I'm assuming that you're using the v4 release of CARTA to test, but just to make sure, please do that (the development version of the frontend has some new changes which will require updates to this code).

This branch doesn't include the general API changes from the other branch (yet, but I'm hoping to merge that branch in soon).

confluence:
The main refactoring PR has now been merged into dev, and dev has been merged into this PR.

kswang1029:
some comments for now:

1. Region objects have some unnecessary attributes and methods. For examples:
- point region: it has attributes and methods as 

> ['CUSTOM_CLASS', 'REGION_TYPES', '__class__', '__delattr__', '__dict__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__', '_base_path', '_region', 'call_action', 'center', 'color', 'control_points', 'dash_length', 'delete', 'existing', 'export_to', 'focus', 'from_list', 'get_value', 'line_width', 'lock', 'macro', 'name', 'new', 'point_shape', 'point_width', 'region_class', 'region_id', 'region_set', 'region_type', 'session', 'set_center', 'set_control_point', 'set_control_points', 'set_line_style', 'set_name', 'set_point_style', 'set_size', 'size', 'unlock', 'wcs_center', 'wcs_size']
	
line_width, set_size, etc are not required for a point region. 

- vector annotation: it has attributes and methods as

> ['CUSTOM_CLASS', 'REGION_TYPES', '__class__', '__delattr__', '__dict__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__', '_base_path', '_region', 'call_action', 'center', 'color', 'control_points', 'dash_length', 'delete', 'endpoints', 'existing', 'export_to', 'focus', 'from_list', 'get_value', 'length', 'line_width', 'lock', 'macro', 'name', 'new', 'pointer_length', 'pointer_width', 'region_class', 'region_id', 'region_set', 'region_type', 'rotation', 'session', 'set_center', 'set_control_point', 'set_control_points', 'set_endpoints', 'set_length', 'set_line_style', 'set_name', 'set_pointer_style', 'set_rotation', 'set_size', 'size', 'unlock', 'wcs_center', 'wcs_endpoints', 'wcs_length', 'wcs_size']

set_size, size etc are not required for a vector annotation.

There are other examples in different regions and annotations. Based on the GUI, I compiled a table for the props of regions/annotations: https://docs.google.com/spreadsheets/d/1XMYA2g1QqzWdm-AQCUlamCKGCiF5819D7ujM5Jv95X0/edit?usp=sharing From there we can see common props are only a few so maybe the base class should only include those. Also from the table, we can see there are groups with common props (*: with additional props):

> 	- point
> 		- point region
> 		- point annotation
> 	- line
> 		- line region
> 		- line annotation
> 		- vector annotation*
> 	- rectangle
> 		- rectangle region
> 		- rectangle annotation
> 		- text annotation*
> 	- ellipse
> 		- ellipse region
> 		- ellipse annotation
> 	- polygon
> 		- polygon region
> 		- polygon annotation
> 		- polyline region
> 		- polyline annotation
> 	- misc
> 		- compass annotation (~vector annotation + text annotation)
> 		- ruler annotation (~line annotation + text annotation)

So my question is, can we have relevant props per region/annotation object (cf GUI)? We may need to reorganize the region class hierarchy?

2. Dual methods for creating some types of regions/annotations

	add_rectangle(center, size, annotation=False, rotation=0, name='')
		- in this case, can we also support the definition with the two diagonal corners (like what we can do via GUI)?
		- in this case (and ellipse), I suggest we arrange the positional arguments as
			add_rectangle(center, size, rotation=0, annotation=False, name='')

	add_vector(start, end, name='')
		- in this case, can we also support the definition with a center, a length, and a position angle?

3. in the GUI we use P.A. as the "rotation" angle for some regions. Shall we consider make it consistent in the two interfaces? In fact, shall we try to make the names of props consistent (eg (py) pointer_length vs (GUI) arrowhead length) in the two interfaces?

confluence:
@kswang1029 I'm going to go through your comments in more detail, but briefly:

* My intention was to provide relevant methods to similar objects, but also a consistent interface between different types -- so that, for example, `size` / `set_size` do something sensible for any object (even if this option is not available in the GUI). So the lack of 1:1 correspondence with the GUI is intentional -- but perhaps some of my choices are confusing (I agree that having a size for point regions is weird -- but does the line width not change the appearance of the point? I will test this). I will provide a more detailed response.
* I treated the ruler as a line-like object, and tried to make its interface consistent with the other line objects.
* It should be easy to add the option to create rectangles and ellipses using corners; I'll look into this.
* I agree that the property names should be consistent; in some cases I followed the names in the frontend source (which don't always match the UI). I will check these.

izkgao:
Some comments and questions:

1. Do we need to unify the inconsistent behaviors on mismatched world / image coordinates for `vector`, `ruler`, and `line`?

`add_line` only allows (image, image) and (world, world), while `add_vector` and `add_ruler` accept (image, world) and (world, image).

```python
# works
vector = regions.add_vector(start=(100, 100), end=('17:56:21.34', '-21:57:23.16'))
# works
ruler = regions.add_ruler(start=(100, 100), end=('17:56:21.34', '-21:57:23.16'))
# does not work
line = regions.add_line(start=(100, 100), end=('17:56:21.34', '-21:57:23.16'))
# ValueError: ('17:56:21.34', '-21:57:23.16') is not a pair of numbers.
```

2. `set_control_point` and `set_control_points` do not accept world coordinates.

```python
vector.set_control_point(1, ('17:56:21.34', '-21:57:23.16'))
# ValueError: ('17:56:21.34', '-21:57:23.16') is not a pair of numbers.
vector.set_control_points([('17:56:21.70', '-21:57:28.16'), ('17:56:21.34', '-21:57:23.16')])
# ValueError: ('17:56:21.70', '-21:57:28.16') is not a pair of numbers.
```

3. Add guards to invalid WCS coordinates

If the returned image coordinates are like (-1.7976931348623157e+308, -1.7976931348623157e+308), an error should be raised before `add_region`.

```python
# Add a point with invalid WCS coordinates
## image.from_world_coordinate_points([("12:12:12.12", "12:12:12.12")])
## [(-1.7976931348623157e+308, -1.7976931348623157e+308)]
point = regions.add_point(center=("12:12:12.12", "12:12:12.12"))

print(point)
# CartaActionFailed: CARTA scripting action .fetchParameter called with parameters (Macro('frameMap[4].regionSet.regionMap[-1]', 'regionType'),) failed: TypeError: Cannot read properties of undefined (reading 'regionType')

print(regions.list())
# CartaActionFailed: CARTA scripting action .fetchParameter called with parameters (Macro('frameMap[4].regionSet.regionMap[-1]', 'regionType'),) failed: TypeError: Cannot read properties of undefined (reading 'regionType')

# A failed region was created with a negative ID.
print(regions.get_value("regionList"))
# [{'id': 0, 'type': 0}, {'id': -1, 'type': 0}]

# This failed region cannot be cleared.
regions.clear()
print(regions.get_value("regionList"))
# [{'id': 0, 'type': 0}, {'id': -1, 'type': 0}]
```

confluence:
@izkgao I've gone through your feedback:

1. This should work now.
2. This is intentional: control points can be interpreted as positions or sizes, depending on the type of the region, so it would be complicated to validate them generically for any region. However, each region type should already also have a more high-level unambiguously-named property with image and wcs getters and a setter that *does* accept values in world coordinates. For example, a vector has `endpoints` (and `wcs_endpoints` and `set_endpoints`), and a polygon has `vertices` (and `wcs_vertices`, etc.).
3. What are the bounds for valid WCS coordinates in the frontend? Is this a global limit, or something that varies between images?

izkgao:
@confluence 

3. I found that creating regions with pixel coordinates exceeding the range of `float32` will fail.

image

izkgao (approved):

confluence:
Per Slack discussion, we should create a separate issue for the bounds problem, and make sure that it's fixed everywhere applicable

kswang1029 (approved):

kswang1029:
@confluence ready for merging.

PR #137, opened 2023-08-16
Simplify distinction between pixel/image units and WCS units

confluence:
This PR simplifies the way the wrapper handles pixel and WCS units, to make it easier to add WCS support to regions while maintaining a consistent API.

* Removed string representation of pixel units (strings like `"100px"` are no longer recognised).
* Bare numbers are now always interpreted as pixel values. Numeric strings with no units are still interpreted as either degrees (if they're coordinates) or arcseconds (if they're angular sizes).
* Removed option to set coordinate system in set_center; this must now be done separately.
* Updated docstrings and tests.
* Number validator no longer matches numeric strings.


kswang1029 (changes requested):
I tested the following cases:
img0.set_center(-50, -50) # working
img0.set_center("7:09:00", "-11:12:37") # working
img0.set_center("07h07m26.5s", "-09d46m57.124s") # working
img0.set_center("108.099deg", "-11.259deg") # not working

Is it possible to support the last use cases as well even if the coordinate system is ICRS/FK5/FK4? Usually we use sexagesimal format to represent coordinates but it is also legal to use decimal format (degree, degree). In some catalogs, source coordinates are listed in decimal format instead of sexagesimal format. So it would be handy to support this use case too.

For the GALACTIC/ECLIPTIC coordinate system, mostly we use decimal format but it is also legal to use sexagesimal format.

confluence:
> Is it possible to support the last use cases as well even if the coordinate system is ICRS/FK5/FK4?

The frontend can only parse world coordinates using the currently set number formats. This doesn't work in the GUI either -- you can't enter degrees in the center dialog if HMS is required. But you *should* be able to change the coordinate system, or set custom number formats without changing the system, and then use the other format. I think that I previously tested both in the wrapper. So in this case, it should be possible to do this:
```
from carta.constants import NumberFormat
session.set_custom_number_format(NumberFormat.DEGREES, NumberFormat.DEGREES)
img0.set_center("108.099deg", "-11.259deg")
```

(But of course then sexagesimal would stop working until you set the formats to sexagesimal or cleared them back to the system default.)

I have found a bug related to this in the GUI: if you have the preferences dialog open, change the custom number formats to something that is not the default for the current system, and then switch to the pan and zoom panel, the format change will not reflect in the inputs for the center (the mouseover format labels are updated, but the numbers are displayed in the wrong format and the new format is not accepted). If you close the preferences dialog and open it again, or if you select a different system (in the pan and zoom dialog or in the global dialog), the updated formats are shown.

kswang1029:
> > Is it possible to support the last use cases as well even if the coordinate system is ICRS/FK5/FK4?
> 
> The frontend can only parse world coordinates using the currently set number formats. This doesn't work in the GUI either -- you can't enter degrees in the center dialog if HMS is required. But you _should_ be able to change the coordinate system, or set custom number formats without changing the system, and then use the other format. I think that I previously tested both in the wrapper. So in this case, it should be possible to do this:
> 
> ```
> from carta.constants import NumberFormat
> session.set_custom_number_format(NumberFormat.DEGREES, NumberFormat.DEGREES)
> img0.set_center("108.099deg", "-11.259deg")
> ```
> 
> (But of course then sexagesimal would stop working until you set the formats to sexagesimal or cleared them back to the system default.)

OK, never mind. I guess users can use astropy or other tools to apply coordinate conversions first then feed the results to set_center.

kswang1029:
> I have found a bug related to this in the GUI: if you have the preferences dialog open, change the custom number formats to something that is not the default for the current system, and then switch to the pan and zoom panel, the format change will not reflect in the inputs for the center (the mouseover format labels are updated, but the numbers are displayed in the wrong format and the new format is not accepted). If you close the preferences dialog and open it again, or if you select a different system (in the pan and zoom dialog or in the global dialog), the updated formats are shown.

thanks for spotting this. issue filed https://github.com/CARTAvis/carta-frontend/issues/2235

kswang1029 (approved):

PR #147, opened 2023-12-29
selenium find_element syntax upgrade

### LINKED ISSUE #144: find_element_by_* are removed starting Selenium 4.3.0
Due to this removed method, the carta python wrapper cannot be run if users are with Selenium 4.3.0+ installed.

The affected codebase is "browser.py" in line90 (https://github.com/CARTAvis/carta-python/blob/c7e7c6c221cc4fc3aa3e4f8b5253ef95fe3ee064/carta/browser.py#L90C17-L90C113):

`session_id = int(self.driver.find_element_by_id("info-session-id").get_attribute("textContent"))`

A solution could be adding
`from selenium.webdriver.common.by import By` in "browser.py" and change line90 to

`session_id = int(self.driver.find_element(By.ID, "info-session-id").get_attribute("textContent"))`

loveluthien:
Fix #144, upgraded selenium find_element syntax. ```find_element_by_id``` is no longer supported after version 4.3. The old function is replaced with ```find_element```.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ This package is not yet published on PyPi, but can be installed from the local r
 
     pip install .
 
-To create a new frontend session which is controlled by the wrapper instead of connecting to an existing frontend session, you also need to install the `selenium` Python library. You also need to make sure that your desired browser is installed, together with a corresponding web driver.
+To create a new frontend session which is controlled by the wrapper instead of connecting to an existing frontend session, you also need to install the `selenium 4.3+` Python library. You also need to make sure that your desired browser is installed, together with a corresponding web driver.
 
 Some example usage of the client as a module is shown in the [documentation](https://carta-python.readthedocs.io).
 
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ This package is not yet published on PyPi, but can be installed from the local r
 
     pip install .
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -7,7 +7,7 @@ This package is not yet published on PyPi, but can be installed from the local r
 
     pip install .
 
-To create a new frontend session which is controlled by the wrapper instead of connecting to an existing frontend session, you also need to install the `selenium` Python library. You also need to make sure that your desired browser is installed, together with a corresponding web driver.
+To create a new frontend session which is controlled by the wrapper instead of connecting to an existing frontend session, you also need to install the `selenium 4.3+` Python library. You also need to make sure that your desired browser is installed, together with a corresponding web driver.
```
confluence: I checked the Selenium Python source, and it looks like [this API has been available for the past 14 years](https://github.com/SeleniumHQ/selenium/commit/d2ffd4b185e6d9d24cfe4051c0723c03a8359606) (it's just the deprecation that's new). So we don't need to require a newer Selenium version here (this code works fine on my currently installed 3.141.0), and we don't need to do anything to support the old API (if we did, we could have done an import test like the one for `StrEnum`).

The actual code change looks good to me.

confluence (changes requested):

README.md
loveluthien: It's good news that we don't have to worry much about backward compatibility.

confluence (approved):

kswang1029 (approved):

PR #161, opened 2026-04-15
Add ColorBlending module for multi-image blending support

### LINKED ISSUE #160: Support color blending
Support color blending

izkgao:
### Description
Closes #160.

Need [frontend PR #2759](https://github.com/CARTAvis/carta-frontend/pull/2759) to be merged first. (merged)

This PR adds color blending support and refactors the session/image APIs around a shared "image-view item" model so frame-backed images and color blendings can be handled consistently.

### What is implemented

- Add a new `ColorBlending` API with `Layer` objects for:
  - listing, adding, replacing, deleting, and reordering layers
  - preserving per-layer alpha during reordering/subsetting
  - setting layer alpha and colormaps
  - toggling raster/contour/vector visibility
- Introduce `ImageBase` as the shared abstraction for image-view items and move common behaviors such as `make_active()` and `image_view_order` onto the base class
- Update `Image` to align with the image-view-item model:
  - use `file_id` terminology consistently
  - expose richer `__repr__` output
  - share activation and image-list-order behavior through `ImageBase`
- Extend `Session` with heterogeneous image-view-item support:
  - `image_list()` now returns both `Image` and `ColorBlending`
  - add `_find_image_view_order()`
  - replace the old image lookup helper with keyword-only `image_by_id(image_view_order=..., file_id=..., color_blending_id=...)`
  - replace `active_frame()` with `active_image()`
  - add `create_color_blending()` and `open_as_color_blending()`
  - add cached `carta_version` and include it in `Session.__repr__` when available
  - update `set_cursor()` to use a better API
- Add new constants for `ImageType` and `ColormapSet`
- Update documentation and quickstart examples to cover:
  - heterogeneous image lists
  - image-view order
  - color blending workflows

**Checklist**
- [x] Add code examples in documentation
- [x] Add tests


docs/source/quickstart.rst
```diff
@@ -225,7 +226,117 @@ Properties which affect the whole session can be set through the session object:
...
+
+    # Make a color blending object
+    # Warning: setting `append=False` will close any existing images
+    # Note: The base layer (index = 0) cannot be deleted or moved.
+    files = [
+        "data/hdf5/first_file.hdf5",
+        "data/fits/second_file.fits",
+        "data/fits/third_file.fits",
+    ]
+    cb = ColorBlending.from_files(session, files, append=False)
```
confluence: This class method should not be exposed to the user directly. The user should access it through a wrapper on the session object, e.g. something like `session.open_as_color_blending([path1, path2, path3])`.

This is analogous to the `open_images` function on the session.

docs/source/quickstart.rst
```diff
@@ -225,7 +226,117 @@ Properties which affect the whole session can be set through the session object:
...
+.. code-block:: python
+
+    from carta.colorblending import ColorBlending
+    from carta.constants import Colormap, ColormapSet
+
+    # Make a color blending object
+    # Warning: This will break the current spatial matching and
+    #          use the first image as the spatial reference
+    # Note: The base layer (index = 0) cannot be deleted or moved.
+    cb = ColorBlending.from_images(session, [img0, img1, img2])
```
confluence: Similarly, this should not be exposed. I'm not sure if it would be best to add the wrapper method to the session object (e.g. maybe `session.color_blending_from([img0, img1, img2])`) or to the image object (to be called on the image that's the spatial reference, e.g. `img0.color_blending_with([img1, img2])`).

docs/source/quickstart.rst
```diff
@@ -225,7 +226,117 @@ Properties which affect the whole session can be set through the session object:
...
+the session helper:
+
+.. code-block:: python
+
+    # Get all open color blending objects in this session
+    color_blendings = session.color_blending_list()
+    cb = color_blendings[0]
+
+    # Or get a color blending object by its image view index
+    cb = ColorBlending.from_imageview_id(session, 3)
```
confluence: This should also not be exposed. Is each colour blending also an image that appears in the image list? If yes, then the wrapper for this should be on the image object, and be decorated as an attribute, since the only parameter is the image's ID. Then the user could access it with `img.color_blending`.

Broader question: if each color blending is also an image and can be accessed through an image object in the wrapper, does that break anything (because it's not a normal image)? E.g. what will happen if the user accesses a blending as an image object and tries to get the number of channels?

If this does break image functionality, I think that the correct thing to do is to specialise `Image` into subclasses (`ImageBase`, `Image`, and `ColorBlending`), but I don't want to overthink this if it's not actually an issue.

docs/source/quickstart.rst
```diff
@@ -225,7 +226,117 @@ Properties which affect the whole session can be set through the session object:
...
+
+.. code-block:: python
+
+    # Get layer objects
+    layers = cb.layer_list()
+
+    # Set colormap for individual layers
+    layers[0].set_colormap(Colormap.REDS)
+    layers[1].set_colormap(Colormap.GREENS)
+    layers[2].set_colormap(Colormap.BLUES)
```
confluence: I suggest unpacking in the assignment, like this:
```
# Get layer objects
red, green, blue = cb.layer_list()

# Set colormap for individual layers
red.set_colormap(Colormap.REDS)
green.set_colormap(Colormap.GREENS)
blue.set_colormap(Colormap.BLUES)
```

carta/colorblending.py
```diff
@@ -0,0 +1,556 @@
...
+    def _validate_initial_layer_count(cls, layer_count):
+        if layer_count > cls.MAX_INITIAL_LAYERS:
+            raise ValueError(
+                "Color blending initialization supports at most "
+                f"{cls.MAX_INITIAL_LAYERS} images (the base layer plus "
+                f"{cls.MAX_INITIAL_LAYERS - 1} matched images)."
+            )
+
+    @classmethod
+    def from_imageview_id(cls, session, imageview_id):
```
confluence: This is the same as the image object ID, right? For consistency, we should call this `image_id`.

carta/colorblending.py
```diff
@@ -0,0 +1,556 @@
...
+        -------
+        :obj:`carta.colorblending.ColorBlending`
+            A new color blending object.
+        """
+        # Find the store ID for the given image view ID
+        path = f"imageViewConfigStore.imageList[{imageview_id}]"
+        image_type = session.get_value(f"{path}.type")
+        if image_type != ImageType.COLOR_BLENDING:
+            raise ValueError(
+                "imageview_id does not refer to a color blending image."
```
confluence: This message should actually include the value of the ID.

carta/colorblending.py
```diff
@@ -0,0 +1,556 @@
...
+
+        Returns
+        -------
+        string
+            The image name.
+        """
+        return self.get_value("filename")
+
+    @property
+    def imageview_id(self):
```
confluence: Similarly, this should be `image_id`.

confluence (changes requested):
I would like the factory classmethods not to be exposed to the user -- they should have wrappers (in the same way that the `Image` factory classmethods are wrapped by methods on the session object). I also have a question about what happens if the user interacts with an image object that corresponds to a color blending (will anything break?).

carta/colorblending.py
izkgao: `imageview_id` refers to the index in the image list (see the red box),
image
while `image_id` in `carta-python` refers to `fileId` in `carta-frontend`, which is not exposed to users and does not change after moving from index 0 to index 2 in the image list.

confluence: Right, I've remembered how this works now. :) As discussed on slack, I think that we should:
* rename the current `image_id` in image objects to `file_id`
* use `image_id` to refer to the image list indices (exposing both in the repr)
* use `blending_id` and `image_id` in the color blending objects
* unify Image and ColorBlending into the same object hierarchy using a shared base class
* retrieve both normal files and color blendings in the image list function in the session
* let functions that refer to an image by ID should accept multiple ID types (as appropriate)
* *not* cache the image ID in the objects (because it can change in the frontend), and fetch it dynamically via the stable internal ID instead

carta/color_blending.py
```diff
@@ -0,0 +1,449 @@
...
+            The alpha values.
+        """
+        layer_list = self.layer_list()
+        if len(alpha_list) != len(layer_list):
+            raise ValueError(
+                f"alpha_list length ({len(alpha_list)}) does not match "
+                f"the number of layers ({len(layer_list)})."
+            )
+        for alpha, layer in zip(alpha_list, layer_list):
+            layer.set_alpha(alpha)
```
confluence: The validation code has features for handling this type of check (`Evaluate` and `Attr` / `Attrs` -- e.g. see how `Image.set_channel` and `Image.set_polarization` are validated).

You would need to create a `@property` for the number of layers (called e.g. `depth`). You could then use `Evaluate` to construct the `IterableOf` at runtime, passing the `Attr` as the minimum and maximum size parameter (they will be evaluated dynamically):
```
@validate(Evaluate(IterableOf, Number(0, 1), Attr("depth"), Attr("depth")))
```

(Untested, but I'm pretty sure that will work.)

You then don't need to do this validation inside. You could use the same `Attr` to validate the layer index in the functions below.

carta/color_blending.py
```diff
@@ -0,0 +1,449 @@
...
+        """Delete a layer from the color blending.
+
+        Parameters
+        ----------
+        layer_index : {0}
+            The layer index. The base layer (layer_index = 0) cannot
+            be deleted.
+        """
+        if layer_index == 0:
+            raise ValueError("The base layer cannot be deleted.")
```
confluence: If the layer index can't be 0, then 0 should be excluded in the validation decorator.

What happens if the provided index is out of range?

You could use the `Evaluate` approach suggested above to use the dynamically evaluated number of layers as the upper bound. The syntax should be exactly the same as the validation for `Image.set_channel`.

carta/color_blending.py
```diff
@@ -0,0 +1,449 @@
...
+
+        Parameters
+        ----------
+        image : {0}
+            The image to set.
+        layer_index : {1}
+            The layer index. The base layer (layer_index = 0) cannot
+            be set.
+        """
+        self.call_action("setSelectedFrame", layer_index - 1, image._frame)
```
confluence: What if the provided index is out of range?

You could use the same validation here.

carta/color_blending.py
confluence: I would additionally provide a function for this on the layer object, which would not require validation, and make it easier for the user to delete a layer if they have already assigned layers to individual objects, e.g.:
```
r, g, b = cb.layer_list()
(...)
# now it's easier to do this:
# g.delete()
# than this:
cb.delete_layer(1) # now you have to remember the index
```

carta/color_blending.py
confluence: I would also additionally provide a function for this on the layer object:
```
r, g, b = cb.layer_list()
(...)
# now it's easier to do this:
# g.set_image(my_img)
```

carta/color_blending.py
```diff
@@ -0,0 +1,449 @@
...
+        layer_index : {0}
+            The layer index. The base layer (layer_index = 0) cannot
+            be deleted.
+        """
+        if layer_index == 0:
+            raise ValueError("The base layer cannot be deleted.")
+        self.call_action("deleteSelectedFrame", layer_index - 1)
+
+    @validate(InstanceOf(Image), Number(1, None))
+    def set_layer(self, image, layer_index):
```
confluence: I would rename this to `set_layer_image`, and switch the parameter order. What is being set is an image; the layer index determines the layer that is being acted on, and the image is the value that is being applied.

carta/color_blending.py
confluence: Would it make sense to be more permissive with the deletion of the base layer, and add special handling?
* If there are other layers, the next lowest layer becomes the base layer
* If there are no other layers, the whole color blending is deleted

carta/session.py
```diff
@@ -507,46 +533,251 @@ def open_hypercube(self, image_paths, append=False):
...
+            elif entry_type == ImageType.COLOR_BLENDING:
+                result.append(ColorBlending(self, entry["id"]))
+            else:
+                raise NotImplementedError(
+                    f"image_list encountered an unsupported image-view "
+                    f"entry at order {order} with type {entry_type!r}; "
+                    "only Image (FRAME) and ColorBlending (COLOR_BLENDING) "
+                    "entries are currently wrapped."
+                )
+        return result
```
confluence: 1. Will this currently crash if there are any open PV previews? It should just skip over them, and print a message.
2. The class selection logic should be implemented in the `ImageBase` class. There's an example of similar logic in the `Region` code: the `CUSTOM_CLASS` dict is a class property on the base class, and its `__init_subclass__` is overridden so that each subclass is automatically added to the dictionary when it is defined. Your use case is simpler: some region subclasses are used for two region types (a region and an annotation), hence the loop, but each of your image subclasses only corresponds to one type (which should be a property on each subclass). I would still implement this logic in the same way instead of hardcoding the dictionary, because this avoids introducing a circular dependency of the base class on the subclasses.

carta/session.py
```diff
@@ -507,46 +533,251 @@ def open_hypercube(self, image_paths, append=False):
...
+            entry_id = entry["id"]
+            if entry_type == ImageType.FRAME:
+                return Image(self, entry_id)
+            if entry_type == ImageType.COLOR_BLENDING:
+                return ColorBlending(self, entry_id)
+            raise NotImplementedError(
+                f"image_by_id encountered an unsupported image-view entry "
+                f"at order {image_view_order} with type {entry_type!r}; "
+                "only Image (FRAME) and ColorBlending (COLOR_BLENDING) "
+                "entries are currently wrapped."
```
confluence: This logic should be in the ImageBase class, as described above.

carta/session.py
```diff
@@ -507,46 +533,251 @@ def open_hypercube(self, image_paths, append=False):
...
+            given_values = {
+                key: val for key, val in provided.items() if val is not None
+            }
+            raise ValueError(
+                "image_by_id requires exactly one of the keyword arguments "
+                "`image_view_order`, `file_id`, or `color_blending_id`; "
+                f"got {len(given)} with values {given_values!r}."
+            )
+
+        summary = self.get_value("imageViewConfigStore.imageListSummary")
```
confluence: Fetching the whole summary here and parsing it in the wrapper is inefficient (especially if you're using the frame or blending ID, in which case you're doing a linear search), especially given that you're calling this in a comprehension in `_images`. My suggestions:

If you're using the order index: 
* use a macro to index the summary *in the frontend* and fetch only the entry at that index: `entry = self.get_value(Macro("imageViewConfigStore",f"imageListSummary[{image_view_order}]"))` (untested, but should work)
* handle the error if the index is out of range
* otherwise construct the object from the type in the summary (moving the class logic into the image base), or fail if it's a PV preview

If you're using the frame ID:
* use a macro to look up the frame in the `frameMap` in the frontend; use the `return_path` option to strip the return value back to the ID (this should only succeed if the frame ID is valid): `image_id = self.get_value(Macro("", f"frameMap[{file_id}]"), return_path="frameInfo.fileId")`
* handle the error if the ID is not found
* otherwise construct the image object

If you're using the blending ID:
* Do the same things as above, but with the color mapping map: `cm_id = self.get_value(Macro("imageViewConfigStore", f"colorBlendingImageMap[{file_id}]"), return_path="id")`

This would greatly simplify this function.

carta/session.py
```diff
@@ -507,46 +533,251 @@ def open_hypercube(self, image_paths, append=False):
...
+            If the active image is of a type that is not yet wrapped on
+            the Python side.
+        """
+        active = self.get_value("activeImage")
+        active_type = active["type"]
+        active_id = active["store"]["id"]
+        if active_type == ImageType.FRAME:
+            return Image(self, active_id)
+        if active_type == ImageType.COLOR_BLENDING:
+            return ColorBlending(self, active_id)
```
confluence: This logic should be in the image base.

carta/wcs_overlay.py
```diff
@@ -437,9 +437,10 @@ class ImageWCSConnector:
 
     def _images(self, image_ids=None):
         """Internal helper function for fetching image objects."""
+        from .image import Image
         if image_ids is None:
-            return self.session.image_list()
-        return [self.session.image_by_id(image_id) for image_id in image_ids]
+            return [img for img in self.session.image_list() if isinstance(img, Image)]
+        return [self.session.image_by_id(file_id=image_id) for image_id in image_ids]
```
confluence: I suggest rewriting this to use the list or map of frames directly instead (to avoid these checks), renaming it to `images` to make it part of the public API, and adding a function to do the same with color mappings (and their list / map).

docs/source/quickstart.rst
```diff
@@ -175,8 +175,46 @@ Helper methods on the session object open images in the frontend and return imag
...
+    from carta.image import Image
+    from carta.color_blending import ColorBlending
+
+    # All open image-view items, in display order
+    items = session.image_list()
+
+    # Filter by type if needed
+    images = [i for i in items if isinstance(i, Image)]
+    color_blendings = [i for i in items if isinstance(i, ColorBlending)]
+
```
confluence, line 215: I think it would be more user-friendly to provide separate helper functions for this -- I have suggested repurposing the existing `_images` (with modifications) and adding a corresponding function for the color mappings.

carta/color_blending.py
izkgao: I noticed that when a user deletes `g`, `b` shifts into the `g` position. The `g` layer would then use the `BLUES` colormap, making the color-based naming misleading. I therefore renamed `r`, `g`, and `b` to `layer1`, `layer2`, and `layer3`, since colormap is an attribute of the underlying image (frame), not a fixed layer position.

carta/session.py
izkgao: It seems that `get_value` accepts a path string rather than a `Macro`. It converts the path into a `Macro` internally before calling `fetchParameter`, so callers should pass the path directly. Explicitly wrapping the path in `Macro` would not work with the current `get_value` implementation.

carta/wcs_overlay.py
izkgao: I added `session.images(self, file_ids=None)` for frames and `session.color_blendings(self, color_blending_ids=None)` for color blending images and call those methods from `Session` for `ImageWCSConnector`.

carta/session.py
```diff
@@ -507,46 +534,312 @@ def open_hypercube(self, image_paths, append=False):
...
+            try:
+                resolved_file_id = self.get_value(
+                    f"frameMap[{file_id}]",
+                    return_path="frameInfo.fileId",
+                )
+            except (CartaActionFailed, CartaBadResponse) as e:
+                raise RuntimeError(
+                    f"No frame-backed image with file_id={file_id} is open."
+                ) from e
+            return ImageBase.image_class(ImageType.FRAME)(
```
confluence: Here you don't need to do this; you can use `Image` directly.

carta/session.py
```diff
@@ -507,46 +534,312 @@ def open_hypercube(self, image_paths, append=False):
...
+                f"imageViewConfigStore.colorBlendingImageMap"
+                f"[{color_blending_id}]",
+                return_path="id",
+            )
+        except (CartaActionFailed, CartaBadResponse) as e:
+            raise RuntimeError(
+                f"No color blending with color_blending_id={color_blending_id} "
+                "is open."
+            ) from e
+        return ImageBase.image_class(ImageType.COLOR_BLENDING)(
```
confluence: Similarly, here you can use ColorBlending directly.

carta/color_blending.py
```diff
@@ -0,0 +1,523 @@
...
+
+            Image(self.session, layers[1].file_id).make_spatial_reference()
+            return
+        self.call_action("deleteSelectedFrame", layer_index - 1)
+
+    @validate(
+        Evaluate(Number, 0, Attr("depth"), Number.INCLUDE_MIN, step=1),
+        InstanceOf(Image),
+    )
+    def set_layer_image(self, layer_index, image):
```
confluence, line 394: What if the image is already set as a different layer? Is there validation in the GUI to prevent that, and should we be adding validation here?

What should happen if the user tries that? I would expect either an error or for the existing layer to be deleted first. But that may be inconvenient, because there's currently no way for the user to reorder layers (should there be? Is there functionality for this in the GUI?).

confluence: I checked, and it looks like there aren't currently any functions for reordering layers in the frontend store. It may be nice to implement them in the wrapper layer (the simplest way would probably be to delete layers and re-add them), but I guess that it wouldn't add much (the user could also just delete and re-add or recreate the whole blending).

carta/image_base.py
confluence, line 1: General naming suggestions:

* Rename `image_view_order` to `index` or `view_index`.

* Use the term "view" or "image view" to describe something that is either a frame or a color blending (or a PV preview) and "image" to describe a frame-backed image only. I think that this is the best solution for resolving the current awkward naming issue (and you're already using "image view" in some descriptions). You can't really call frames "files" because you might have a concatenated hypercube or a LEL expression (so not a single file), but "frame" feels like an overly technical implementation term.
 
If you use "view" as the general description, you can rename session's `image_list` to `views`, which I think will make more sense combined with `images` and `color_blendings`. You could then also rename `ImageBase` to `View` or `ImageView`. `file_id` could go back to `image_id`.

confluence (commented):
I'm happy with these changes; I have a few more minor suggestions and a question.

carta/image_base.py
izkgao: Great idea!

carta/color_blending.py
izkgao: Yes, that's a good point. I checked the current behavior and found that if the image is already in another layer, calling `set_layer_image()` with `layer_index == 0` removes the existing base layer and makes the selected image the new spatial reference. As a result, the total layer depth decreases by one. That is not what I would expect from `set_layer_image()`. It should replace the image in a layer without changing the number of layers. I've added validation to reject images that are already present to prevent that from happening. To make creating a new color blending more convenient, `Session.create_color_blending()` now accepts an ordered, non-repeating list of `Image` objects and creates the layers in that order.

carta/session.py
```diff
@@ -510,43 +537,360 @@ def open_hypercube(self, image_paths, append=False):
...
+        return [
+            view_class(
+                self,
+                self.get_value(
+                    f"{path}[{index}]",
+                    return_path=return_path,
+                ),
+            )
+            for index in range(count)
+        ]
```
confluence: It's not necessary to use multiple `get_value` calls to fetch the items from these lists one by one.

`images` could have the same implementation as the old `image_list` -- you can call `get_value` once to get `frameNames` (but we probably don't really need `Image.from_list`; that logic can be done inline).

There's no equivalent serializable list of color blending IDs in the stores, but in `color_blendings` you could use `imageViewConfigStore.imageListSummary` like you do in `views` (with a comprehension to get the color blendings only and create objects from them).

You could actually do that in `images` as well, for consistency. I guess that it's slightly more efficient to filter the view type in the frontend than in the wrapper because you're sending less data, but this is very small data, so it's not going to be a big difference.

_Ideally, the image view store should have an action that takes an ImageType parameter and returns a list of IDs of views of that type -- you could then use that to get exactly what you need for these functions. But that can be a later enhancement; the existing frontend properties are fine for now._

_In future it would also be helpful if the frontend's scripting service could apply the return path to all the values in a list or map if that is the return value -- then we would be able to query existing object collections from the wrapper directly, and wouldn't have to add serializable filters for them. There may be other features to add at the same time; e.g. using a list or map of paths for `return_path` to get a list or map of values back._

confluence (commented):
One more issue with fetching lists -- this is the last thing; everything else looks good!

carta/session.py
izkgao: I've added two frontend issues for what you suggested and will open another `carta-python` PR for a light refactor once the two corresponding frontend PRs are merged.
https://github.com/CARTAvis/carta-frontend/issues/2820
https://github.com/CARTAvis/carta-frontend/issues/2903

carta/session.py
```diff
@@ -652,16 +671,28 @@ def color_blendings(self, color_blending_ids=None):
...
+                ColorBlending(self, entry["id"])
+                for entry in summary
+                if entry["type"] == ImageType.COLOR_BLENDING
+            ]
+
+        color_blending_ids_in_summary = {
+            entry["id"]
+            for entry in summary
+            if entry["type"] == ImageType.COLOR_BLENDING
+        }
```
confluence: Minor issue: you could eliminate the duplication here by creating the filtered list of IDs at the start (but only converting it to a set later, so that you have the ordering for all entries).

carta/session.py
```diff
@@ -630,8 +640,17 @@ def images(self, image_ids=None):
             The requested images.
         """
         if image_ids is None:
-            return self._view_list("frames", Image, "frameInfo.fileId")
-        return [self.view_by_id(image_id=image_id) for image_id in image_ids]
+            return [
+                Image(self, entry["value"]) for entry in self.get_value("frameNames")
+            ]
+
+        frame_ids = {entry["value"] for entry in self.get_value("frameNames")}
```
confluence: You could similarly remove the duplication here (get the list of IDs first; convert to a set here).

carta/session.py
```diff
@@ -630,8 +640,17 @@ def images(self, image_ids=None):
...
+        images = []
+        for image_id in image_ids:
+            if image_id not in frame_ids:
+                raise RuntimeError(f"No image with image_id={image_id} is open.")
+            images.append(Image(self, image_id))
```
confluence: Here you could subtract the set of all IDs from the set of requested IDs to get a set of missing IDs, and raise an error listing all the missing IDs. Then you could use a comprehension here as well to create and return the list of objects.

carta/session.py
```diff
@@ -652,16 +671,28 @@ def color_blendings(self, color_blending_ids=None):
...
+        }
+        color_blendings = []
+        for color_blending_id in color_blending_ids:
+            if color_blending_id not in color_blending_ids_in_summary:
+                raise RuntimeError(
+                    f"No color blending with "
+                    f"color_blending_id={color_blending_id} is open."
+                )
+            color_blendings.append(ColorBlending(self, color_blending_id))
+        return color_blendings
```
confluence: Same remark as for the other function: you could use set arithmetic to find the missing IDs and then a comprehension to create the list.

carta/session.py
```diff
@@ -574,44 +574,54 @@ def views(self, view_indices=None):
...
-                print(
-                    f"Skipping unsupported view entry at index {index}: "
-                    f"{entry!r}: {e}"
-                )
+                result.append(View.view_class(entry["type"])(self, entry["id"]))
+            except (CartaValidationFailed, NotImplementedError):
+                if not skip_unsupported:
+                    raise
+                view_type = ImageType(entry["type"])
+                print(f"Skipping unsupported {view_type.name} view at index {index}.")
```
confluence: I would suggest making this implementation more consistent with the others -- rather than checking if any indices are larger than the length of the list, you could use the same set logic to find missing indices.

You'd just have the extra step of checking for unsupported types and possibly skipping them, so the construction should be done in a full loop like it is here and not in a comprehension.

carta/session.py
confluence: I've realised that there's a way to reorder this logic that is much cleaner and avoids repetition (applicable to all three functions):
* all_image_ids = [comprehension to extract IDs from frameNames]
* If no image_ids:
  * image_ids = all_image_ids
* Else:
  * missing_image_ids = set(image_ids) - set(all_image_ids)
  * If missing_image_ids: raise error
* return [comprehension to make objects from image_ids]

confluence: (In `views` the last bit needs to be a loop, so that you can handle unsupported types.)

confluence (commented):
One more set of comments about the list code.

Also, a general comment about code style: because comprehensions are supposed to be compact, I favour using short iterator variable names rather than the long descriptive names that would be more appropriate in a loop. So e.g. it would be fine to write `[i for i in some_list if i > something]`.

(If you prefer to use long names everywhere because of a linting tool, that's fine.)

carta/session.py
izkgao: Brilliant! I love this. This saves a lot of lines and makes it easier to read. Thanks.

I use a list to obtain the missing ids instead to keep them in the original order for the error message.

confluence (approved):
I'm happy to merge this with @kswang1029's approval. I will review the remaining PRs once they're updated after the merge.

carta/session.py
```diff
@@ -510,43 +534,328 @@ def open_hypercube(self, image_paths, append=False):
         image_id = self.call_action(command, stokes_images, output_directory, output_hdu)
         return Image(self, image_id)
 
-    def image_list(self):
```
confluence, line 513: I've realised that we should deprecate this method instead of removing it outright, otherwise existing scripts will break. We should:
* print a deprecation warning telling people to use `images` instead (or `views` for all views or `color_blendings`...)
* call `images` (with no IDs)
* update the description to add a deprecation warning.

It's fine not to do this for `Images.from_list`, because that wasn't part of the public API (the description clearly said not to use it directly).

carta/session.py
```diff
@@ -510,43 +534,328 @@ def open_hypercube(self, image_paths, append=False):
...
+
+    @validate(NoneOr(IterableOf(Number.ID)))
+    def views(self, view_indices=None):
+        """Return all or selected currently open views.
+
+        When no indices are supplied, all open views are returned. When
+        indices are supplied, the views at those positions are returned in
+        the requested order.
 
-    def active_frame(self):
```
confluence, line 523: This also needs to be deprecated in a similar way -- I suggest calling `active_view` and throwing an exception if it's not an image.

carta/session.py
```diff
@@ -510,43 +534,328 @@ def open_hypercube(self, image_paths, append=False):
...
+            If an explicitly requested view type is unsupported.
+        CartaValidationFailed
+            If ``view_indices`` contains an invalid value.
         """
-        image_id = self.get_value("activeFrame.frameInfo.fileId")
-        return Image(self, image_id)
+        summary = self.get_value("imageViewConfigStore.imageListSummary")
+        remote_indices = range(len(summary))
 
-    def image_by_id(self, image_id):
```
confluence, line 534: And so does this. This is straightforward; we can call `view_by_id` with the image ID.

confluence (changes requested):
Sorry to backtrack, but there are three session functions that we should initially deprecate instead of removing, to avoid breaking existing scripts.

confluence (approved):
That's great; thanks!

PR #169, opened 2026-04-30
Validate enums

### LINKED ISSUE #157: Revise colormap code
The current colormap constants exclude the special monochromatic presets that the frontend accepts. It would be good to remove these hardcoded values entirely, and to validate colormap strings against a list obtained dynamically from the frontend (similarly to the way we validate image dimension parameters). The list should also be made available to the user.

izkgao:
## Description

Closes #157.
Need frontend https://github.com/CARTAvis/carta-frontend/pull/2759 to be merged first.

Adds enum snapshot checks to keep Python constants aligned with frontend/protobuf enums.

## What is implemented

- Align constants with frontend/protobuf enum snapshots and add private enum mismatch diagnostics with an opt-in live test.

## How to test
1. Build this frontend branch https://github.com/CARTAvis/carta-frontend/pull/2759 or after
2. Launch CARTA with `--enable_scripting`
3. Replace the url and session id below with yours.
```bash
CARTA_RUN_LIVE_ENUM_SNAPSHOTS=1 \
CARTA_FRONTEND_URL='http://127.0.0.1:3002' \
CARTA_SESSION_ID='1584862638' \
uv run pytest tests/live/test_live_enum_snapshots.py
```
4. See only this error
Because `ANNULUS` has not been implemented yet, we do not add it to `RegionType` in `constants.py`.
```python
AssertionError: [{'missing_in_frontend': [], 'missing_in_python': [{'frontend_name': 'ANNULUS', 'frontend_value': 5}], 'name': 'RegionType', 'source': 'protobuf:RegionType', ...}]
assert [{'missing_in...onType', ...}] == []
```


confluence (commented):
I'm going to wait until the other PRs have been resolved before doing a final review, but I have an alternative suggestion for specifying the links to protobuf and frontend enums:

* Create mix-in classes called `FrontendEnum` and `ProtobufEnum`.
* For each of these, implement a registration mechanism for subclasses (similar to the region and view mechanism), which:
  * Checks the subclass for an *optional* `snapshot_name` class attribute (maybe this should be called something like `external_name` or `reference_name` -- and it should probably be in all caps for style consistency)
  * Updates a class dict that maps snapshot names to enum subclasses (if no custom `snapshot_name` is provided, the name of the class is used by default)
* Make each class that is linked to a frontend or protobuf enum inherit one of these mixins.

Now by default enums won't be linked to anything, and you can access the mappings of the mixins to get the lists of frontend- and protobuf-linked enums (instead of having to tag every enum and parse them all manually), and you don't have to provide external names explicitly unless the class name is different.

izkgao:
I removed the unrelated code changes from `zhenkai/check_version` branch so it can be reviewed independently.

izkgao:
I found that it would be better to make it work without launching CARTA. I will make a script in frontend to generate a list of enum and use it for validation.

